PR #73, opened 2024-03-13
Warp addressed transactions

aaronbuchwald:
# ACP-73: Warp Addressed Transactions

[Rendered](https://github.com/aaronbuchwald/ACPs/blob/warp-addressed-transactions/ACPs/73-warp-addressed-transactions.md)

## Abstract

Add a standard for `Warp Addressed Transactions` to enable any `(blockchainID, address)` pair to sign a a transaction that is executable on any destination chain. This is intended as a general standard that can be implemented by any VM on the Avalanche Network. As the first instance, this ACP proposes integrating Warp Addressed Transactions to the P-Chain to enable Subnet Driven Validator Sets as first laid out in https://github.com/avalanche-foundation/ACPs/discussions/68.

## Diagram

To illustrate the full flow, we provide a sequence graph showing the issuance of a Warp Addressed Transaction from Account1 on BlockchainA to BlockchainB. Note that there are no restrictions on whether BlockchainA and BlockchainB are the same, different, on the same Subnet, or the Primary Network. They only need to agree on the standard for Warp Addressed Transactions.

To keep the diagram compact, we use the following abbreviations:

* `WAT` - Warp Addressed Transaction
* `WAS` - Warp Addressed Signature

```mermaid
flowchart TB
    A[Account1 on BlockchainA] -->|1. Call Warp with WAT Payload| B[BlockchainA]
    B -->|2. Accept Account1 Tx and Generate WAS| C[Relayer]
    C -->|3. Issue WAT to BlockchainB| D[BlockchainB]
```

Note: if the Warp Addressed Transaction can be issued entirely on its own, the relayer can be replaced by sending a cross-chain message over the network to the destination chain.

## Motivation

Subnets are currently created and orchestrated entirely on the Avalanche P-Chain. Many Subnets may have the desire or need to maintain greater sovereignty over their validator set, reward curve, and overall tokenomics than what is available on the P-Chain.

Warp Addressed Transactions enable transferring ownership of a Subnet to a `(blockchainID, address)` pair, which can be an arbitary program on any chain in the Avalanche Ecosystem (ex. manage your validator set from a contract on your own chain or a contract on the C-Chain).

By transferring ownership to a Warp Derived Address, Subnets can manage their validator set against a generic API.

In the immediate term, this accomplishes two critical goals for Subnets:

1. Enable Token Standard PoS Subnets (ERC-20 Staking)
2. Enable generic validator set and tokenomics management

Further, this standard could be implemented by any VM that wants to enable Warp Addressed Transactions.

PR #75, opened 2024-04-05
Consensus Proofs

joshua-kim:


joshua-kim:
Images seem to look strange but I believe they will work once merged (preview): https://github.com/joshua-kim/ACPs/blob/consensus-proofs/ACPs/75_consensus_proofs/75-consensus-proofs.md

aaronbuchwald:
Repeating @richardpringle 's suggestion on ACP-73: https://github.com/avalanche-foundation/ACPs/pull/73#issuecomment-1999866840, it would be great to add a link to the branch/relevant file to the PR description, so you can quickly jump to it.

ACPs/75_consensus_proofs/75-consensus-proofs.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+The `Container` message is used to send a container with its corresponding
+consensus proof.
+
+```diff
++ message Container {
++   bytes container = 1;
++   bytes proof = 2;
```
patrick-ogrady, line 74: Do we want to be more specific about what we think a `proof` could be here (i.e. a BLS Multi-Signature)? Otherwise, I think "Consensus Proof" could be interpreted a lot of different ways.

ACPs/75_consensus_proofs/75-consensus-proofs.md
```diff
@@ -0,0 +1,87 @@
...
+### P2P
+
+The `Ancestors` message is replaced with the `containers`
+field which will supply a set of containers with their corresponding proofs.
+
+```diff
+message Ancestors {
+    bytes chain_id = 1;
+    uint32 request_id = 2;
+-   repeated bytes containers = 3;
```
patrick-ogrady, line 63: nit: mark `3` as reserved

ACPs/75_consensus_proofs/75-consensus-proofs.md
```diff
@@ -0,0 +1,87 @@
...
+
+If ProposerVM blocks were produced with a proof of the P-Chain block to be verified against instead of just its height, verifiers could continue making progress by requesting any block they had not accepted yet from another peer and verifying their corresponding acceptance proofs. If a block’s proof is valid, the blocks can be executed in-order locally to finally verify the proposed subnet block.
```
patrick-ogrady, line 22: Do we want to more explicitly callout that this could also be used by non-validating nodes to accept blocks instead of running consensus (which would mean many less connections + les load on validators)?

ACPs/75_consensus_proofs/75-consensus-proofs.md
```diff
@@ -0,0 +1,87 @@
...
++   bytes container = 1;
++   bytes proof = 2;
++ }
+```
+
+## Security Considerations
+
+Nodes that bootstrap using state sync may not have the entire history of the
+P-Chain and therefore will not be able to provide the entire history for a block
+that referenced in a block they propose. This would be needed to unblock a node that is attempting to fast-forward their P-Chain, as they require the entire ancestry between their current accepted tip and the block they are attempting to forward to. This is considered to be acceptable as
```
patrick-ogrady, line 82: nit: `that is referenced`

ACPs/75_consensus_proofs/75-consensus-proofs.md
```diff
@@ -0,0 +1,87 @@
...
+nodes will be expected to maintain some minimum of some recent state. An alternative is to make a churn assumption and validate the proposed block's proof with a stale validator set.
```
patrick-ogrady, line 83: May also want to note that this is `ok` because it is assumed that `some` node has the data and with enough retries they will eventually get it? (unless we want to talk about how nodes should only keep recent blocks).

ACPs/75_consensus_proofs/75-consensus-proofs.md
```diff
@@ -0,0 +1,87 @@
...
+
+The block header will require an update to include a Proof of the parent block
+to be verified at corresponding to the existing height field.
+
+```diff
+type BlockHeader struct {
+    PChainHeight uint64
+    Timestamp    time.Time
+    Proposer     ids.NodeID
++   Proof        []byte
```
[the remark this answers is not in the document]
joshua-kim: This is pseudocode just intended to be used as a reference so if it's in a separate struct/not I don't have a strong opinion on. Regarding a delay - I believe this does need to be on a delay as I think it needs to be the proof for the block at `PChainHeight` which is currently already on a delay via window-ing.

ACPs/75_consensus_proofs/75-consensus-proofs.md
joshua-kim: Yeah I can add this as a section.

richardpringle:
@joshua-kim, you `Rendered` link broke with the file rename. The new link should be here:
https://github.com/joshua-kim/ACPs/blob/consensus-proofs/ACPs/75_consensus_proofs/75-acceptance-proofs.md

ACPs/75_consensus_proofs/75-acceptance-proofs.md
```diff
@@ -0,0 +1,90 @@
...
+}
+```
+
+The `Container` message is used to send a container with its corresponding
+acceptance proof.
+
+```diff
++ message Container {
++   bytes container = 1;
++   bytes bls_multi_signature = 2;
```
aaronbuchwald, line 77: Should this be a warp signature (with bitset) or a direct multi-signature?

ACPs/75_consensus_proofs/75-acceptance-proofs.md
```diff
@@ -0,0 +1,90 @@
...
++ }
+```
+
+## Security Considerations
+
+Nodes that bootstrap using state sync may not have the entire history of the
+P-Chain and therefore will not be able to provide the entire history for a block
+that is referenced in a block they propose. This would be needed to unblock a node that is attempting to fast-forward their P-Chain, as they require the entire ancestry between their current accepted tip and the block they are attempting to forward to. This is considered to be acceptable as
+nodes will be expected to maintain some minimum of some recent state. An alternative is to make a churn assumption and validate the proposed block's proof with a stale validator set.
```
aaronbuchwald, line 86: Could we rephrase "some minimum of som recent state" ?

aaronbuchwald (commented):
Great work, added some style comments

ACPs/75_consensus_proofs/75-acceptance-proofs.md
```diff
@@ -0,0 +1,90 @@
...
+![figure 3](./3.jpg)
+
+## Specification
+
+Note: The following is pseudocode and is provided only as visual aid.
+
+### ProposerVM
+
+The block header will require an update to include a Proof of the parent block
+to be verified at corresponding to the existing height field.
```
michaelkaplan13, line 45: The wording of this sentence is confusing to me. 

The `Proof` is a proof of acceptance of the parent block on the given chain right? Not a proof of acceptance of the `PChainHeight` block on the P-Chain?

joshua-kim: Yeah this was poorly worded. Rephrased.

PR #77, opened 2024-04-16
placeholder

dhrubabasu:


ACPs/13-subnet-only-validators.md
```diff
@@ -3,8 +3,9 @@ ACP: 13
 Title: Subnet-Only Validators (SOVs)
 Author(s): Patrick O'Grady <contact@patrickogrady.xyz>
 Discussions-To: https://github.com/avalanche-foundation/ACPs/discussions/14
-Status: Proposed
+Status: Stale
 Track: Standards
+Superseded-By: 13
```
patrick-ogrady, line 8: I agree that ACP-77 should supersede ACP-13 👍

sofianeOuafir:
Really like the ideas in this ACP! It's going to make subnet validation a lot more flexible and open up plenty of new possibilities. Super excited to see new chains use the P-chain for interoperability with other chains in the Avalanche ecosystem. This feels like a big step towards what a network of networks should look like. 👏

GuruTanuki:
Why is it mandatory to abolish the current system if this ACP is accepted ? 

That way the current subnets would have the choice to do it or not, it would give free will to the subnet that has trusted Avalanche up to that point. 

We would have two systems. 

One subscription format and another one with no fees but with a big initial bet. Any opinions ?

Thank you for this ACP good job.

0xsergen:
I believe that launching on Avalanche with this proposal will be a no-brainer choice for projects. By the way, I think it should be considered to give some portion of the subnet fees to the P-chain validators to encourage validating the P-chain.

atinst:
Please proceed, as time waits for no one.

PR #101, opened 2024-06-06
clarify SetSubnetValidatorManagerTx

ceyonur:
It's a bit hard to follow the difference between `SourceAddress` and `Address`. We should clarify that these are two different fields in the warp message with `AddressedCall` payload

### DIFF
--- a/ACPs/77-reinventing-subnets/README.md
+++ b/ACPs/77-reinventing-subnets/README.md
@@ -157,7 +157,7 @@ The `Message` field in the above transaction must be an Avalanche Warp Message u
                        +----------+
 ```
 
-- `messageID` is the sha256 hash of the warp message in the `RegisterSubnetValidatorTx` that registered the Subnet Validator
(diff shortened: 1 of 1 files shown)


ACPs/77-reinventing-subnets/README.md
```diff
@@ -192,7 +194,8 @@ type SetSubnetValidatorManagerTx struct {
     // Warp message should include:
     //   - SubnetID
     //   - ChainID (where the validator manager lives)
```
ceyonur, line 196: Is this same with `SourceChainID` or is that also a separate field in `AddressedCall`? cc @dhrubabasu

dhrubabasu:
All the `// Warp message should include:` comments are the payload for the `AddressedCall`. The `SourceAddress` isn't listed there. Definitely could be clearer, thanks for looking at it!

ceyonur:
We've removed `SetSubnetValidatorManagerTx` and fixed those in https://github.com/avalanche-foundation/ACPs/pull/135

PR #103, opened 2024-06-07
[ACP-103] Add Dynamic Fees to the X-Chain and P-Chain

dhrubabasu:
Introduce a dynamic fee mechanism to the X-Chain and the P-Chain. Preview a future transition to a multidimensional fee mechanism.

This ACP is an alternative to [ACP-83](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/83-dynamic-multidimensional-fees/README.md) with two notable changes:
* Simpler fee mechanism
* Synthesizing the multi-dimensional parameters into a single "gas" parameter for single dimension fees

### DIFF
--- a/ACPs/103-dynamic-fees/README.md
+++ b/ACPs/103-dynamic-fees/README.md
@@ -13,7 +13,7 @@ Introduce a dynamic fee mechanism to the X-Chain and the P-Chain. Preview a futu
 
 ## Motivation
 
-Blockchains are computation-limited environments. Users are charged for the execution and inclusion of their transactions based on the blockchain's transaction fee mechanism. The mechanism should fluctuate based on supply and demand dynamics to serve as a deterrent against spam and denial-of-service attacks.
+Blockchains are resource-constrained environments. Users are charged for the execution and inclusion of their transactions based on the blockchain's transaction fee mechanism. The mechanism should fluctuate based on supply of and demand for said resources to serve as a deterrent against spam and denial-of-service attacks.
 
 With a fixed fee mechanism, users are provided with simplicity and predictability but network congestion and resource constraints are not taken into account. There is no incentive for users to withhold transactions since the cost is fixed regardless of the demand. The fee does not adjust the execution and inclusion fee of transactions to a market clearing price.
 
@@ -124,7 +124,7 @@ A full reference implementation has not been provided yet. It must be provided p
 
 ## Security Considerations
 
-The current fixed fee mechanism on the X-Chain and P-Chain does not robustly handle spikes in load. Switching these chains to a dynamic fee mechanism will stabilize the load over the long-term.
+The current fixed fee mechanism on the X-Chain and P-Chain does not robustly handle spikes in load. Switching these chains to a dynamic fee mechanism will ensure that load is properly priced given allotted processing capacity.
 
 ## Acknowledgements
 


ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+1. `Bandwidth` ($B$) is the amount of network bandwidth used in transaction broadcast. This is set to the size of the transaction in bytes.
+2. `Reads` ($R$) is the number of state/database reads used in transaction execution.
+3. `Writes` ($W$) is the number of state/database writes used in transaction execution.
+4. `Compute` ($C$) is the total amount of compute used in transaction execution. This is set to the number of signatures verified during transaction execution.
+
+For each transaction, gas ($G$) can be computed (TODO: The weights for each dimension must be specified prior to this ACP being considered "Implementable"):
+
+$$G = B + R + W + C$$
+
+A future ACP should remove the gas standardization to granularly meter usage of each resource in a multidimensional scheme.
```
StephenButtolph, line 37:
```suggestion
A future ACP could remove the gas standardization to granularly meter usage of each resource in a multidimensional scheme.
```

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+- $\exp\left(x\right)$ is an approximation of $e^x$
+- $K$ is a constant to control the rate of change for gas price
+
+After processing block $b$, $x$ is updated with the total gas in the block $G$:
+
+$$x = x + G$$
+
+Whenever $x$ increases by $K$, the gas price increases by a factor of `~2.7`. If the gas price gets too expensive, average usage drops, and $x$ starts decreasing, automatically dropping the price again. The gas price constantly adjusts to make sure that, on average, the blockchain uses $T$ gas per second.
+
+$L$ is defined as the maximum gas per second that the blockchain has capacity for. For monotonically increasing blocktimes (as on P-Chain and X-Chain), there may be multiple blocks issued at the same on-chain timestamp. $L$ limits the amount of gas that can be used at $t$, placing a bound on the amount that the required fee can increase from $t$ to $t'$. Once the chain has used $L$ amount of gas at time $t$, a new block will only be considered valid if its timestamp is at least $t + 1$.
```
StephenButtolph, line 65: I'm not sure that this is really the mechanism we had talked about (offline). I was under the impression we would be specifying a moving capacity, not a capacity at a specific timestamp.

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+$$b_{n+1} = b_n \cdot e^x$$
+
+Assume $b_n = 100$ and the current block is 1 unit above target utilization, or $x = 1$. Then, $b_{n+1} = 100 \cdot e^1 \approx 271.828$, an increase of `271%`. If instead $b_n = 10,000$, $b_{n+1} = 10,000 \cdot e^1 \approx 27,182.8$, an increase of `271%` again. The fee is _equally_ reactive as the fee increases. This is because the rate of change scales with $x$: $\frac{d}{dx}e^x = e^x$.
+
+### Block Building Procedure
+
+When a transaction is constructed on the X-Chain and P-Chain, the amount of $AVAX burned in a transaction is given by `sum($AVAX outputs) - sum($AVAX inputs)`. The amount of gas used by the transaction can also be deterministically calculated after construction. Dividing the amount of $AVAX burned by the amount of gas yields the maximum gas price that the transaction can pay.
+
+Instead of using a FIFO queue for the mempool (like the X-Chain and P-Chain do now), the mempool should be a priority queue ordered by the maximum gas price of each transaction. This ensures that higher paying transactions are included first.
```
StephenButtolph, line 109:
```suggestion
Instead of using a FIFO queue for the mempool (like the X-Chain and P-Chain do now), the mempool should be a priority queue ordered by the gas price of each transaction. This ensures that higher paying transactions are included first.
```

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Backwards Compatibility
+
+Modfication of a fee mechanism is an execution change and requires a mandatory upgrade for activation. Implementers must take care to not alter the execution behavior prior to activation.
+
+After this ACP is activated, any transaction issued on the P-Chain or X-Chain must account for the fee mechanism defined above. There is a concern of underpriced transactions filling the mempool and causing memory issues on nodes. To alleviate memory pressure, the mempool should be periodically cleared of underpriced transactions. Users are responsible for re-broadcasting underpriced transactions when the fee lowers or re-constructing their transaction to include a larger fee.
```
StephenButtolph, line 115: I'm not sure I understand this concern, the mempools should have a maximum size and the a tx paying a higher fee should be able to evict a transaction paying a lower fee.

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
+```text
+ACP: 103
+Title: Add Dynamic Fees to the X-Chain and P-Chain
+Author(s): Dhruba Basu (https://github.com/dhrubabasu), Stephen Buttolph (https//github.com/StephenButtolph), Alberto Benegiamo (https://github.com/abi87)
```
StephenButtolph, line 4: If we're going alphabetical by last name:
```suggestion
Author(s): Dhruba Basu (https://github.com/dhrubabasu), Alberto Benegiamo (https://github.com/abi87), Stephen Buttolph (https//github.com/StephenButtolph)
```

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+- $\exp\left(x\right)$ is an approximation of $e^x$
+- $K$ is a constant to control the rate of change for gas price
+
+After processing block $b$, $x$ is updated with the total gas in the block $G$:
+
+$$x = x + G$$
+
+Whenever $x$ increases by $K$, the gas price increases by a factor of `~2.7`. If the gas price gets too expensive, average usage drops, and $x$ starts decreasing, automatically dropping the price again. The gas price constantly adjusts to make sure that, on average, the blockchain uses $T$ gas per second.
+
+$L$ is defined as the maximum gas per second that the blockchain has capacity for. For monotonically increasing blocktimes (as on P-Chain and X-Chain), there may be multiple blocks issued at the same on-chain timestamp. $L$ limits the amount of gas that can be used at $t$, placing a bound on the amount that the required fee can increase from $t$ to $t'$. Once the chain has used $L$ amount of gas at time $t$, a new block will only be considered valid if its timestamp is at least $t + 1$.
```
abi87, line 65: Chain time precision for X and P chain is 1 second already. Can we simplify this and define L as max complexity per block?

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+The initial parameters for the P-Chain will be set to:
+
+| Parameter | Value |
+| - | - |
+| $M$ - minimum gas price | TODO |
+| $T$ - target gas per second | TODO |
+| $L$ - max gas per second | TODO |
+| $K$ - change constant | TODO |
+
+As the network gains capacity to handle additional load (through the activation of future protocol enhancements), this algorithm can be tuned to increase the number of transactions that can be processed at a specific fee rate.
```
abi87, line 85: We don't strictly need protocol enhancements to improve capacity. Even a better state handling may without protocol changes may get us there

abi87: I'd mention gas or complexity processed instead of number of transactions. $L$ does not cap the number of transactions

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+At the start of building/executing block $b$, $x$ is updated:
+
+$$x = \max(x - (T \cdot \Delta t), 0)$$
+
+Where $\Delta t$ is the number of seconds between $b$ and $b$'s parent block.
+
+The required fee per gas for block $b$ is:
+
+$$M \cdot \exp\left(\frac{x}{K}\right)$$
```
abi87, line 51: can we introduce a notation for "fee rate applied to block b"?
We reuse $b_n$ in a section below with a difference meaning, which does not help readability

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Let's use a linear adjustment function:
+
+$$b_{n+1} = b_n + 10x$$
+
+Assume $b_n = 100$ and the current block is 1 unit above target utilization, or $x = 1$. Then, $b_{n+1} = 100 + 10 \cdot 1 = 110$, an increase of `10%`. If instead $b_n = 10,000$, $b_{n+1} = 10,000 + 10 \cdot 1 = 10,010$, an increase of `0.00001%`. The fee is _less_ reactive as the fee increases. This is because the rate of change is constant: $\frac{d}{dx}10x = 10$.
+
+Now, let's use an exponential adjustment function:
+
+$$b_{n+1} = b_n \cdot e^x$$
```
abi87, line 101: q: wouldn't any multiplicative rule work here?
$$b_{n+1} = b_n \cdot f(x)$$

ACPs/103-dynamic-fees/README.md
dhrubabasu: adjusted the phrasing

abi87 (changes requested):

ACPs/103-dynamic-fees/README.md
dhrubabasu: It cannot be per block because we can issue an infinite amount of complexity at time $t$. Reworking the section to be a moving capacity target based on time

ACPs/103-dynamic-fees/README.md
dhrubabasu: $e$'s the simplest exponential function for theory purposes. We'll likely be using $2^x$ in the code

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+For each transaction, gas ($G$) can be computed (TODO: The weights for each dimension must be specified prior to this ACP being considered "Implementable"):
+
+$$G = B + R + W + C$$
+
+A future ACP could remove the gas standardization to granularly meter usage of each resource in a multidimensional scheme.
+
+### Mechanism
+
+This mechanism aims to have a target gas usage ($T$) per second and adjusts based on the excess gas usage ($x$), defined as the difference between the current gas usage and $T$.
```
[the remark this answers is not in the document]
dhrubabasu: I think it looks better in prose to have it in parenthesis 🤷🏾‍♂️

ACPs/103-dynamic-fees/README.md
abi87: I don't see how we get to infinite complexity. Consecutive blocks can have same timestamp which means that cumulated complexity will go up with the full block complexity (without being reduced by $T \cdot \Delta t$.
$L$ cap can be set on a per block base and each subsequent block will see the its fee increased

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+1. `Bandwidth` ($B$) is the amount of network bandwidth used in transaction broadcast. This is set to the size of the transaction in bytes.
+2. `Reads` ($R$) is the number of state/database reads used in transaction execution.
+3. `Writes` ($W$) is the number of state/database writes used in transaction execution.
+4. `Compute` ($C$) is the total amount of compute used in transaction execution. This is set to the number of signatures verified during transaction execution.
+
+For each transaction, gas ($G$) can be computed (TODO: The weights for each dimension must be specified prior to this ACP being considered "Implementable"):
+
+$$G = B + R + W + C$$
+
+A future ACP could remove the gas standardization to granularly meter usage of each resource in a multidimensional scheme.
```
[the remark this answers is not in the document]
dhrubabasu: The coefficients won't be changed in a multi-dimensional scheme. Each resource will have an individual fee market and capacity

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+The C-Chain, in [Apricot Phase 3](https://medium.com/avalancheavax/apricot-phase-three-c-chain-dynamic-fees-432d32d67b60), employs a dynamic fee mechanism to raise the price during periods of high demand and lowering the price during periods of low demand. As the price gets too expensive, network utilization will decrease, which drops the price.
+
+The X-Chain and P-Chain currently operate under a fixed fee mechanism. To better account for increased load that may result from [ACP-77](../77-reinventing-subnets/README.md), they should be migrated to a dynamic fee mechanism.
+
+## Specification
+
+### Dimensions
+
+There are four dimensions that will be used to approximate the computational cost ("gas" usage) of a transaction:
```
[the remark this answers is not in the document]
dhrubabasu: "gas" is pretty widely used in the blockchain space and the multidimensional literature for fees: https://arxiv.org/abs/2208.07919

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+A future ACP could remove the gas standardization to granularly meter usage of each resource in a multidimensional scheme.
+
+### Mechanism
+
+This mechanism aims to have a target gas usage ($T$) per second and adjusts based on the excess gas usage ($x$), defined as the difference between the current gas usage and $T$.
+
+At the start of building/executing block $b$, $x$ is updated:
+
+$$x = \max(x - (T \cdot \Delta t), 0)$$
```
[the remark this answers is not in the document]
dhrubabasu: I personally find the parenthesis easier to read

ACPs/103-dynamic-fees/README.md
dhrubabasu: We need both a block capacity cap and a moving capacity cap. A block capacity block by itself won't limit the amount of gas that can be used at $t$ since there can be consecutive blocks with the same timestamp. This could make the fee go up indefinitely without being reduced by $T$.

abi87: Fees can't go up indefinitely since avax has capped supply right? I don't see why we need both capacities. I think the block one should be enough.

ACPs/103-dynamic-fees/README.md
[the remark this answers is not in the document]
aaronbuchwald: I agree with @abi87 this is a bit hard to read. This uses gas standardization for the first time, which is a bit strange.

I'd suggest re-phrasing the part above and here to break this into two parts:

1. Track the amount of each resource consumed (compute/reads/writes/bandwidth)
2. A cost vector for each resource

For ACP-103, the cost vector will be a fixed value for each resource ie. 1 unit of compute is 100 gas and 1 read is 50 gas, etc multiplied by a single base fee.

In the follow up ACP, the cost vector will be updated to a dynamic cost vector where each resource maintains its own base fee.

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+After this ACP is activated, any transaction issued on the P-Chain or X-Chain must account for the fee mechanism defined above. Users are responsible for re-constructing their transaction to include a larger fee for quicker inclusion when the fee increases.
+
+## Reference Implementation
+
+A full reference implementation has not been provided yet. It must be provided prior to this ACP being considered "Implementable".
+
+## Security Considerations
+
+The existing fixed fee mechanism on the X-Chain and P-Chain have led to periods of instability when larger transactions are accepted. Metering these transactions according to their computational cost or "gas" usage will remove those periods of instability at the cost of higher fees for users.
```
aaronbuchwald, line 122: I'd add this will result in higher fees for users during times of congestion and most likely lower fees for users most of the time (depending on how we set target utilization)

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Block Building Procedure
+
+When a transaction is constructed on the X-Chain and P-Chain, the amount of $AVAX burned in a transaction is given by `sum($AVAX outputs) - sum($AVAX inputs)`. The amount of gas used by the transaction can also be deterministically calculated after construction. Dividing the amount of $AVAX burned by the amount of gas yields the maximum gas price that the transaction can pay.
+
+Instead of using a FIFO queue for the mempool (like the X-Chain and P-Chain do now), the mempool should be a priority queue ordered by the maximum gas price of each transaction. This ensures that higher paying transactions are included first.
+
+## Backwards Compatibility
+
+Modfication of a fee mechanism is an execution change and requires a mandatory upgrade for activation. Implementers must take care to not alter the execution behavior prior to activation.
```
aaronbuchwald, line 112:
```suggestion
Modification of a fee mechanism is an execution change and requires a mandatory upgrade for activation. Implementers must take care to not alter the execution behavior prior to activation.
```

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+$$b_{n+1} = b_n \cdot e^x$$
+
+Assume $b_n = 100$ and the current block is 1 unit above target utilization, or $x = 1$. Then, $b_{n+1} = 100 \cdot e^1 \approx 271.828$, an increase of `271%`. If instead $b_n = 10,000$, $b_{n+1} = 10,000 \cdot e^1 \approx 27,182.8$, an increase of `271%` again. The fee is _equally_ reactive as the fee increases. This is because the rate of change scales with $x$: $\frac{d}{dx}e^x = e^x$.
+
+### Block Building Procedure
+
+When a transaction is constructed on the X-Chain and P-Chain, the amount of $AVAX burned in a transaction is given by `sum($AVAX outputs) - sum($AVAX inputs)`. The amount of gas used by the transaction can also be deterministically calculated after construction. Dividing the amount of $AVAX burned by the amount of gas yields the maximum gas price that the transaction can pay.
+
+Instead of using a FIFO queue for the mempool (like the X-Chain and P-Chain do now), the mempool should be a priority queue ordered by the maximum gas price of each transaction. This ensures that higher paying transactions are included first.
```
aaronbuchwald, line 108: How should block builders handle the case that the highest priority transaction does not fit in the block?

ACPs/103-dynamic-fees/README.md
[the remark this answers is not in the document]
aaronbuchwald: Could this specify how rounding will be handled?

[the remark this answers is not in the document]
aaronbuchwald: Could we also use a more descriptive variable name than x?

ACPs/103-dynamic-fees/README.md
dhrubabasu: adjusted the phrasing

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+Where $\Delta t$ is the number of seconds between $b$ and $b$'s parent block.
+
+The required fee per gas for block $b$ is:
+
+$$M \cdot \exp\left(\frac{x}{K}\right)$$
+
+Where:
+
+- $M$ is the minimum gas price
```
aaronbuchwald, line 55: Should this be the previous price instead of the minimum price?

I think the update rule needs to replace the minimum price with the previous price and then apply the minimum.

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Abstract
+
+Introduce a dynamic fee mechanism to the X-Chain and the P-Chain. Preview a future transition to a multidimensional fee mechanism.
+
+## Motivation
+
+Under a fixed fee mechanism, there is a greater likelihood of higher network congestion. During periods of high demand, there is no economic incentive for users to withhold issuance of their transactions, extending the period of high demand.
+
+The C-Chain, in [Apricot Phase 3](https://medium.com/avalancheavax/apricot-phase-three-c-chain-dynamic-fees-432d32d67b60), employs a dynamic fee mechanism to raise the price during periods of high demand and lowering the price during periods of low demand. As the price gets too expensive, network utilization will decrease, which drops the price.
```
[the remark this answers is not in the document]
dhrubabasu: I think calling out to EIP-1559 is a bit strange in this ACP since the C-Chain dynamic fee is only mentioned in passing

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+The X-Chain and P-Chain currently operate under a fixed fee mechanism. To better account for increased load that may result from [ACP-77](../77-reinventing-subnets/README.md), they should be migrated to a dynamic fee mechanism.
```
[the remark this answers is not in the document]
dhrubabasu: adjusted the wording

ACPs/103-dynamic-fees/README.md
dhrubabasu: The block must be partially packed if a transaction does not fit in it. In that case, we'd issue the block and the highest priority transaction would be the first transaction in the next block.

ACPs/103-dynamic-fees/README.md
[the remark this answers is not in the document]
dhrubabasu: > Could this specify how rounding will be handled?

There's no rounding here

> Could we also use a more descriptive variable name than x?

Got any suggestions haha?

ACPs/103-dynamic-fees/README.md
dhrubabasu: I generally avoid introducing notation that I don't use. $b_n$ below is not the same as here (and why I didn't use originally).

I'm not sure where the readability concern is, $b_n$ isn't used here and is only defined below

ACPs/103-dynamic-fees/README.md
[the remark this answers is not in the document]
dhrubabasu: I think that's a bit more detail than I'd like to put in here. I agree that gas standardization is a bit strange. How does this sound instead?

```suggestion
A future ACP could remove the merging of these dimensions and granularly meter usage of each resource in a multidimensional scheme.
```

ACPs/103-dynamic-fees/README.md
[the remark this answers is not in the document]
aaronbuchwald: excessGas ?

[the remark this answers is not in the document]
dhrubabasu: I'd like to keep it to one letter since it's easier to see the form of the update. It's strange to only expand $x$ and not $T$ or $\Delta t$ as well

ACPs/103-dynamic-fees/README.md
dhrubabasu: Updated the ACP to use a moving capacity cap only, a block capacity cap can be implicitly derived from it.

A block cap is not sufficient because block production is asynchronous in Snowman. The article introducing C-Chain dynamic fees has good detail on it: https://medium.com/avalancheavax/apricot-phase-three-c-chain-dynamic-fees-432d32d67b60

> Because block production is asynchronous in Snowman, Moderato cannot depend on a height-based mechanism to determine if utilization is above or below the target (many blocks could be produced in a short window to game the fee mechanism) nor can it rely on the content of blocks alone (there is additional overhead to splitting the same content in 20 blocks instead of 1). To satisfy these constraints, Moderato forms a view of the network using a rolling, time-based window of activity, where block production is reduced into the same measure of activity as the block content itself.

We're using a different window in this ACP to only track remaining capacity over a time window.

ACPs/103-dynamic-fees/README.md
dhrubabasu: Good call- gonna note down that this phrasing should be updated once we have the real fees. Don't want to say "likely" since we're going to know soon

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Introduce a dynamic fee mechanism to the X-Chain and the P-Chain. Preview a future transition to a multidimensional fee mechanism.
+
+## Motivation
+
+Under a fixed fee mechanism, there is a greater likelihood of higher network congestion. During periods of high demand, there is no economic incentive for users to withhold issuance of their transactions, extending the period of high demand.
+
+The C-Chain, in [Apricot Phase 3](https://medium.com/avalancheavax/apricot-phase-three-c-chain-dynamic-fees-432d32d67b60), employs a dynamic fee mechanism to raise the price during periods of high demand and lowering the price during periods of low demand. As the price gets too expensive, network utilization will decrease, which drops the price.
+
+The X-Chain and P-Chain currently operate under a fixed fee mechanism. To better handle increased load, they should be migrated to a dynamic fee mechanism. This migration should be done in tandem with [ACP-77](../77-reinventing-subnets/README.md) to ensure developers can account for the dynamic fee mechanism.
```
StephenButtolph, line 20: Why are we brining up ACP 77 here? This ACP is unrelated to 77 afaict (77 should bring up this ACP as it will use this mechanism for it's fee)

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+## Specification
+
+### Dimensions
+
+There are four dimensions that will be used to approximate the computational cost ("gas" usage) of a transaction:
+
+1. `Bandwidth` ($B$) is the amount of network bandwidth used in transaction broadcast. This is set to the size of the transaction in bytes.
+2. `Reads` ($R$) is the number of state/database reads used in transaction execution.
+3. `Writes` ($W$) is the number of state/database writes used in transaction execution.
+4. `Compute` ($C$) is the total amount of compute used in transaction execution. This is set to the number of signatures verified during transaction execution.
```
StephenButtolph, line 31: Is this going to explicitly be the number of signatures? Or do we just thing that sig verification will dominate this value?
```suggestion
4. `Compute` ($C$) is the total amount of compute used in transaction execution.
```

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Parameter | X-Chain | P-Chain |
+| - | - | - |
+| $T$ - target gas usage per second | TODO | TODO |
+| $M$ - minimum gas price | TODO | TODO |
+| $K$ - gas price update constant | TODO | TODO |
+| $L$ - gas limit constant | TODO | TODO |
+| $S$ - gas limit time period | TODO | TODO |
+
+As the network gains capacity to handle additional load, this algorithm can be tuned to increase the amount of gas that can be processed at a specific fee rate.
```
StephenButtolph, line 87:
```suggestion
As the network gains capacity to handle additional load, this algorithm can be tuned to increase the rate that gas can be processed.
```

ACPs/103-dynamic-fees/README.md
dhrubabasu: good point- removed

ACPs/103-dynamic-fees/README.md
dhrubabasu: adjusted the wording

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Whenever $x$ increases by $K$, the gas price increases by a factor of `~2.7`. If the gas price gets too expensive, average usage drops, and $x$ starts decreasing, automatically dropping the price again. The gas price constantly adjusts to make sure that, on average, the blockchain consumes $T$ gas per second.
+
+A gas limit constant $L$ is defined to limit the gas that can be consumed every $S$ seconds, bounding the amount that the gas fee can increase in $S$ seconds. A remaining gas capacity variable $r$ is defined to track the gas amount remaining for new blocks. At the beginning of processing block $b$, $r$ is set:
+
+$$r = \max\left(r + \frac{L \cdot \Delta{t}}{S}, L\right)$$
+
+Where $\Delta t$ is the number of seconds between $b$ and $b$'s parent block. The maximum gas consumed in $\Delta{t}$ is $r + \Delta{t} \cdot \frac{L}{S}$. The upper bound across all $\Delta{t}$ is $L + \Delta{t} \cdot \frac{L}{S}$.
+
+After $b$ is processed, the total gas consumed in $b$, or $G$, will be known. If $r < G$, $b$ is considered invalid. If $b$ is valid, $r$ is updated:
```
dhrubabasu, line 71:
```suggestion
After $b$ is processed, the total gas consumed in $b$, or $G$, will be known. If $G > r$, $b$ is considered invalid. If $b$ is valid, $r$ is updated:
```

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+Track: Standards
+```
+
+## Abstract
+
+Introduce a dynamic fee mechanism to the X-Chain and the P-Chain. Preview a future transition to a multidimensional fee mechanism.
+
+## Motivation
+
+Under a fixed fee mechanism, there is a greater likelihood of higher network congestion. During periods of high demand, there is no economic incentive for users to withhold issuance of their transactions, extending the period of high demand.
```
patrick-ogrady, line 16: > Under a fixed fee mechanism, there is a greater likelihood of higher network congestion.
I only think of escalating fees in the context of both block supply + demand (not just demand). I think "greater likelihood of network congestion" implies this but doesn't satisfy me (sounds like some game of chance).

> During periods of high demand, there is no economic incentive for users to withhold issuance of their transactions, extending the period of high demand.

This implies that "high demand" is the problem. I think that is the wrong mentality. The problem is that the network does not adjust the fee for inclusion to a market clearing price at capacity (both leaving money on the table for the network and leading users to compete over the protocol stack for inclusion).

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The C-Chain, in [Apricot Phase 3](https://medium.com/avalancheavax/apricot-phase-three-c-chain-dynamic-fees-432d32d67b60), employs a dynamic fee mechanism to raise the price during periods of high demand and lowering the price during periods of low demand. As the price gets too expensive, network utilization will decrease, which drops the price.
+
+The X-Chain and P-Chain currently operate under a fixed fee mechanism. To better handle increased load, they should be migrated to a dynamic fee mechanism.
```
patrick-ogrady, line 20: > To better handle increased load

"to more robustly handle spikes in load"

ACPs/103-dynamic-fees/README.md
abi87: We use $b$ here to indicate a block and $b_n$ below to indicate the fee rates associated with the block.
The two notation are too similat to me.
Please introduce the fee rates notation here and reuse it later. Something like
```
fee_rate = $$M \cdot \exp\left(\frac{x}{K}\right)$$
```

ACPs/103-dynamic-fees/README.md
abi87: The question was: why not using something that is not exponential? Say:
```
feeRate_{n+1} = feeRate_n \times x^2
```
You'd have something "equally reactive" as well.
I don't think we have explained why we need the exponential here

ACPs/103-dynamic-fees/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Where $\Delta t$ is the number of seconds between $b$ and $b$'s parent block.
+
+The required fee per for block $b$ is:
+
+$$M \cdot \exp\left(\frac{x}{K}\right)$$
+
+Where:
+
+- $M$ is the minimum gas price
```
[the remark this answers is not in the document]
dhrubabasu: I think we only use gas price. Where do you see fee rate?

ACPs/103-dynamic-fees/README.md
```diff
@@ -122,7 +124,7 @@ A full reference implementation has not been provided yet. It must be provided p
 
 ## Security Considerations
 
-The existing fixed fee mechanism on the X-Chain and P-Chain have led to periods of instability when larger transactions are accepted. Metering these transactions according to their computational cost or "gas" usage will remove those periods of instability at the cost of higher fees for users.
+The current fixed fee mechanism on the X-Chain and P-Chain does not robustly handle spikes in load. Switching these chains to a dynamic fee mechanism will stabilize the load over the long-term.
```
patrick-ogrady, line 127: > will stabilize the load over the long-term

will ensure that load is properly priced given allotted processing capacity.

****

I don't think this is capable of "stabilizing load" (nor do I think that is the goal).

ACPs/103-dynamic-fees/README.md
```diff
@@ -13,11 +13,13 @@ Introduce a dynamic fee mechanism to the X-Chain and the P-Chain. Preview a futu
 
 ## Motivation
 
-Under a fixed fee mechanism, there is a greater likelihood of higher network congestion. During periods of high demand, there is no economic incentive for users to withhold issuance of their transactions, extending the period of high demand.
+Blockchains are computation-limited environments. Users are charged for the execution and inclusion of their transactions based on the blockchain's transaction fee mechanism. The mechanism should fluctuate based on supply and demand dynamics to serve as a deterrent against spam and denial-of-service attacks.
```
patrick-ogrady, line 16: > computation-limited

resource-constrained

patrick-ogrady: > based on supply and demand dynamics

based on the supply of and demand for said resources.

PR #108, opened 2024-06-11
Avalanche EVM Event Importing Standard (Best Practices)

michaelkaplan13:
Defines a standard smart contract interface and abstract implementation for importing EVM events from any blockchain within Avalanche using [Avalanche Warp Messaging](https://docs.avax.network/build/cross-chain/awm/overview).

### DIFF
--- a/ACPs/108-evm-event-importing/README.md
+++ b/ACPs/108-evm-event-importing/README.md
@@ -1,0 +1,117 @@
+```text
+ACP: 108
+Title: EVM Event Importing Standard
+Author(s): Michael Kaplan (https://github.com/mkaplan13)
+Discussions-To: https://github.com/avalanche-foundation/ACPs/discussions/
+Status: Proposed
+Track: Best Practices Track
+```
+
+## Abstract
+
+Defines a standard smart contract interface and abstract implementation for importing EVM events from any blockchain within Avalanche using [Avalanche Warp Messaging](https://docs.avax.network/build/cross-chain/awm/overview).
+
+## Motivation
+
+The implementation of Avalanche Warp Messaging within `coreth` and `subnet-evm` exposes a [mechanism for getting authenticated hashes of blocks](https://github.com/ava-labs/subnet-evm/blob/master/contracts/contracts/interfaces/IWarpMessenger.sol#L43) that have been accepted on blockchains within Avalanche. There is currently no clear standard for using authenticated block hashes in smart contracts within Avalanche, making it difficult to build applications that leverage this mechanism. In order to make effective use of authenticated block hashes, contracts must be provided encoded block headers that match the authenticated block hashes and also Merkle proofs that are verified against the state or receipts root contained in the block header. 
+
+With a standard interface and abstract contract implemetation that handles the authentication of block hashes and verification of Merkle proofs, smart contract developers on Avalanche will be able to much more easily create applications that leverage data from other Avalanche blockchains. These type of cross-chain application do not require any direct interaction on the source chain.
+
+## Specification
+
+### Event Importing Interface
+
+We propose that smart contracts importing EVM events emitted by other blockchains within Avalanche implement the following interface.
+
+#### Methods
+
+Imports the EVM event uniquely identified by the source blockchain ID, block header, transaction index, and log index. The event data is provided in the `receiptProof`, which must be a valid Merkle proof of the receipt for the given `txIndex` against the receipt root of the `blockHeader`. Must emit an `EventImported` event upon success. Should validate that the `blockHeader` matches an authenticated block hash from the `sourceBlockchainID`.
+```solidity
+function importEvent(
+    bytes32 sourceBlockchainID,
+    bytes calldata blockHeader,
+    uint256 txIndex,
+    bytes[] calldata receiptProof,
+    uint256 logIndex
+) external;
+```
+
+This interface does not require that the Warp precompile is used to authenticate block hashes. Implementations could:
+- Use an external block hash verification mechanism, such as trusted oracles.
+- Use the Warp precompile to authenticate block hashes provided in the transaction calling `importEvent`.
+- Look up previously authenticated block hashes from another contract, such that multiple events could be imported from the same block hash, which would only need to be authenticated once.
+
+#### Events
+
+Must trigger when an EVM event is imported.
+```solidity
+event EventImported(
+    bytes32 indexed sourceBlockchainID,
+    bytes32 indexed sourceBlockHash,
+    address indexed loggerAddress,
+    uint256 txIndex,
+    uint256 logIndex
+);
+```
+
+### Event Importing Abstract Contract
+
+Applications importing EVM events emitted by other blockchains within Avalanche should be able to use a standard abstract implementation of the `importEvent` interface. This abstract implementation should handle:
+- Authenticating block hashes from other chains using Warp precompile.
+- Verifying that the encoded `blockHeader` matches the imported block hash.
+- Verifying the Merkle `receiptProof` for the given `txIndex` against the receipt root of the provided `blockHeader`.
+- Decoding the event log identified by `logIndex` from the receipt obtained from verifying the `receiptProof`.
+
+Inheriting contracts should only need to define the logic to be executed when an event is imported. This is done by providing an implementation of the following internal function, called by `importEvent`.
+
+```solidity
+function _onEventImport(EVMEventInfo memory eventInfo) internal virtual;
+```
+
+Where the `EVMEventInfo` struct is defined as:
+
+```solidity
+struct EVMLog {
+    address loggerAddress;
+    bytes32[] topics;
+    bytes data;
+}
+
+struct EVMEventInfo {
+    bytes32 blockchainID;
+    uint256 blockNumber;
+    uint256 txIndex;
+    uint256 logIndex;
+    EVMLog log;
+}
+```
+
+## Reference Implementation
+
+See reference implementation on [Github here](https://github.com/ava-labs/event-importer-poc).
+
+In addition to implementing the interface and abstract contract described above, the reference implementation shows how transactions can be constructed to import events using Warp block hash signatures.
+
+## Open Questions
+
+See [here](https://github.com/ava-labs/event-importer-poc?tab=readme-ov-file#open-questions-and-considerations).
+
+## Security Considerations
+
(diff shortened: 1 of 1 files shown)


ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Open Questions
+
+See [here](https://github.com/ava-labs/event-importer-poc?tab=readme-ov-file#open-questions-and-considerations).
+
+## Security Considerations
+
+The correctness of a contract using block hashes to prove that a specific event was emitted within that block depends on the correctness of:
+1. The mechanism for authenticating that a block hash was finalized on another blockchain.
+2. The Merkle proof validation library used to prove that a specific transactoin receipt was included in the given block.
```
patrick-ogrady, line 103: nit: `transactoin` -> `transaction`

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Specification
+
+### Event Importing Interface
+
+We propose that smart contracts importing EVM events emitted by other blockchains within Avalanche implement the following interface.
+
+#### Methods
+
+Imports the EVM event uniquely identified by the source blockchain ID, block header, transaction index, and log index. The event data is provided in the `receiptProof`, which must be a valid Merkle proof of the receipt for the given `txIndex` against the receipt root of the `blockHeader`. Must emit an `EventImported` event upon success. Should validate that the `blockHeader` matches an authenticated block hash from the `sourceBlockchainID`.
```
patrick-ogrady, line 28: nit: it may be nice to include a link to the spec for evm block headers?

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+    bytes32 indexed sourceBlockHash,
+    address indexed loggerAddress,
+    uint256 txIndex,
+    uint256 logIndex
+);
+```
+
+### Event Importing Abstract Contract
+
+Applications importing EVM events emitted by other blockchains within Avalanche should be able to use a standard abstract implementation of the `importEvent` interface. This abstract implementation should handle:
```
patrick-ogrady, line 59: nit: IMO it is weird to use `should handle: authenticating block hashes using warp precompile` given you say `does not require the warp precompile` earlier?

That to me is a `could` with the others being `must`?

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+    uint256 blockNumber;
+    uint256 txIndex;
+    uint256 logIndex;
+    EVMLog log;
+}
+```
+
+## Reference Implementation
+
+See reference implementation on [Github here](https://github.com/ava-labs/event-importer-poc).
```
patrick-ogrady, line 91: ❤️

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+Track: Best Practices Track
+```
+
+## Abstract
+
+Defines a standard smart contract interface and abstract implementation for importing EVM events from any blockchain within Avalanche using [Avalanche Warp Messaging](https://docs.avax.network/build/cross-chain/awm/overview).
+
+## Motivation
+
+The implementation of Avalanche Warp Messaging within `coreth` and `subnet-evm` exposes a [mechanism for getting authenticated hashes of blocks](https://github.com/ava-labs/subnet-evm/blob/master/contracts/contracts/interfaces/IWarpMessenger.sol#L43) that have been accepted on blockchains within Avalanche. There is currently no clear standard for using authenticated block hashes in smart contracts within Avalanche, making it difficult to build applications that leverage this mechanism. In order to make effective use of authenticated block hashes, contracts must be provided encoded block headers that match the authenticated block hashes and also Merkle proofs that are verified against the state or receipts root contained in the block header.
```
patrick-ogrady, line 16: nit: you may also want to link to ACP-75 (saying this is a specific instance of why they are useful)

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### Methods
+
+Imports the EVM event uniquely identified by the source blockchain ID, block header, transaction index, and log index. The event data is provided in the `receiptProof`, which must be a valid Merkle proof of the receipt for the given `txIndex` against the receipt root of the `blockHeader`. Must emit an `EventImported` event upon success. Should validate that the `blockHeader` matches an authenticated block hash from the `sourceBlockchainID`.
+```solidity
+function importEvent(
+    bytes32 sourceBlockchainID,
+    bytes calldata blockHeader,
+    uint256 txIndex,
+    bytes[] calldata receiptProof,
```
patrick-ogrady, line 34: nit: may also be nice to link to the `receipt` struct (so they know what they get)

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+We propose that smart contracts importing EVM events emitted by other blockchains within Avalanche implement the following interface.
+
+#### Methods
+
+Imports the EVM event uniquely identified by the source blockchain ID, block header, transaction index, and log index. The event data is provided in the `receiptProof`, which must be a valid Merkle proof of the receipt for the given `txIndex` against the receipt root of the `blockHeader`. Must emit an `EventImported` event upon success. Should validate that the `blockHeader` matches an authenticated block hash from the `sourceBlockchainID`.
+```solidity
+function importEvent(
+    bytes32 sourceBlockchainID,
+    bytes calldata blockHeader,
+    uint256 txIndex,
```
patrick-ogrady, line 33: nit:  not sure if it feels cleaner with `txIndex` after `receiptProof`

ACPs/108-evm-event-importing/README.md
michaelkaplan13: Good call, added now. 

Question though...ACP-75 doesn't directly support surfacing the acceptance within the VM right? Only in consensus?

ACPs/108-evm-event-importing/README.md
michaelkaplan13: I considered it. My rationale for putting `txIndex` first is that the RLP encoded `txIndex` is the key in the receipts trie that the `receiptProof` proves inclusion of (that's obviously not clear from the interface itself though).

michaelkaplan13: I added an explanation of this in the section above now. Let me know if you still think it should be re-ordered

ACPs/108-evm-event-importing/README.md
michaelkaplan13: Added

ACPs/108-evm-event-importing/README.md
michaelkaplan13: Thanks for calling this out. I tweaked the interface to include the explicit `sourceBlockchainID` param after typing the initial draft, and agree this part was left unclear/inconsistent. Updated the wording now to `must`, but removed the "using the Warp precompile" qualifier and added an explanation of how implementations `could` authenticate the block hashes below.

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,117 @@
...
+    bytes[] calldata receiptProof,
+    uint256 logIndex
+) external;
+```
+
+This interface does not require that the Warp precompile is used to authenticate block hashes. Implementations could:
+- Use an external block hash verification mechanism, such as trusted oracles.
+- Use the Warp precompile to authenticate block hashes provided in the transaction calling `importEvent`.
+- Look up previously authenticated block hashes from another contract, such that multiple events could be imported from the same block hash, which would only need to be authenticated once.
```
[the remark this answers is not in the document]
michaelkaplan13: Think it may be best to move this topic to a discussions thread once this PR is merged, but:

Looking at the `importEvents` interface:
```
function importEvent(
    bytes32 sourceBlockchainID,
    bytes calldata blockHeader,
    uint256 txIndex,
    bytes[] calldata receiptProof,
    uint256 logIndex
) external;
```

The pieces that seem EVM specific are the `txIndex` and `logIndex`. The `txIndex` is actually the key of trie that the `receiptProof` proves inclusion of, so maybe that should be renamed `trieKey` (and possibly made `bytes`?). The `logIndex` is extra EVM specific information for knowing how to interpret the data imported using the proof, so that could also be made more generalized. 

I still need to think on it more, but right now I think a reasonable set of assumptions made by the interface would be:
- The hash imported from the `sourceBlockchainID` is the hash of an encoded struct, and the `blockHeader` is the pre-image of that hash.
- The `blockHeader` includes a root hash of a trie.
- The `txIndex` is the key of that trie that `receiptProof` proves the inclusion of.

The tweaks to the interface to reduce it that set of assumptions would probably be:
- Renaming a few params
- Making `txIndex` bytes
- Moving `logIndex` to an extra data field (?)

Curious what your thoughts here are.

ACPs/108-evm-event-importing/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Methods
+
+Imports the EVM event uniquely identified by the source blockchain ID, block header, transaction index, and log index.
+
+The `blockHeader` must be validated to match the authenticated block hash from the `sourceBlockchainID`. The specification for EVM block headers can be found [here](https://github.com/ava-labs/subnet-evm/blob/master/core/types/block.go#L73).
+
+The `txIndex` identifies the key of receipts trie of the given block header that the `receiptProof` must prove inclusion of. The value obtained by verifying the `receiptProof` for that key is the encoded transaction receipt. The specification for EVM transaction receipts can be found [here](https://github.com/ava-labs/subnet-evm/blob/master/core/types/receipt.go#L62).
+
+The `logIndex` identidies which event log from the given transaction receipt is to be imported.
```
meeshavax, line 34: nit: 'identidies' should be 'identifies'

ACPs/108-evm-event-importing/README.md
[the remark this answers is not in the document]
aaronbuchwald: This makes sense to me.

It may be that if we try to support arbitrary VMs, this just turns into a router that can call on a specific implementation or verifier. I'm not sure the resulting data (the receipts in this case), would look the same for every VM. If the inputs and outputs both look different, I'm not sure yet if putting a router in the middle will make sense or whether we should just have different implementations that do not follow the same interface as needed.

aaronbuchwald (approved):

PR #118, opened 2024-07-08
P2P Warp signature request

cam-schultz:
```text
ACP: 118
Title: Warp Signature Interface Standard
Author(s): Cam Schultz (https://github.com/cam-schultz)
Discussions-To: 
Status: Proposed
Track: Standards Track
```

## Abstract

Proposes handling Warp signature requests from peers at the protocol level, rather than leaving it to VMs to implement. This will standardize the signature request interface, allowing for more flexible, VM-agnostic signature aggregation implementations.

[Rendered](https://github.com/cam-schultz/ACPs/tree/warp-signature-request/ACPs/118-warp-signature-request)

### DIFF
--- a/ACPs/118-warp-signature-request/README.md
+++ b/ACPs/118-warp-signature-request/README.md
@@ -1,0 +1,81 @@
+```text
+ACP: 118
+Title: Warp Signature Interface Standard
+Author(s): Cam Schultz (https://github.com/cam-schultz)
+Discussions-To: 
+Status: Proposed
+Track: Standards Track
+```
+
+## Abstract
+
+Proposes handling Warp signature requests from peers at the protocol level, rather than leaving it to VMs to implement. This will standardize the signature request interface, allowing for more flexible, VM-agnostic signature aggregation implementations.
+
+## Motivation
+
+Warp message signatures consist of an aggregate BLS signature composed of the individual signatures of a subnet's validators. Individual signatures need to be retreivable by the party that wishes to construct an aggregate signature. At present, this is left to VMs to implement, as is the case with [Subnet EVM](https://github.com/ava-labs/subnet-evm/blob/v0.6.6/plugin/evm/message/signature_request.go#20) and [Coreth](https://github.com/ava-labs/coreth/blob/v0.13.6-rc.0/plugin/evm/message/signature_request.go#L20)
+
+This creates friction in applications that are intended to operate across many VMs (or distinct implementations of the same VM). As an example, the reference Warp message relayer implementation, [awm-relayer](https://github.com/ava-labs/awm-relayer), fetches individual signatures from validators and aggregates them before sending the Warp message to its destination chain for verification. However, Subnet EVM and Coreth have distinct codecs, requiring the relayer to [switch](https://github.com/ava-labs/awm-relayer/blob/v1.4.0-rc.0/relayer/application_relayer.go#L372) according to the target codebase.
+
+Another example is ACP-75, which aims to implement acceptance proofs using Warp. The signature aggregation mechanism is not [specified](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/75-acceptance-proofs/README.md#signature-aggregation), which is a blocker for that ACP to be marked implementable.
+
+Standardizing the Warp Signature Request format by implementing it as a protocol-level (AvalancheGo) AppRequest type would simplify the implementation of ACP-75, and streamline signature aggregation for out-of-protocol services such as Warp message relayers.
+
+## Specification
+
+Subnet EVM and Coreth currently handle Warp message signature requests from peers by decoding the `AppRequest` bytes (forwarded from Avalanchego via the `AppHandler` interface) using a type registered with the message codec:
+
+```
+type MessageSignatureRequest struct {
+	MessageID ids.ID `serialize:"true"`
+}
+```
+
+`MessageID` represents the VM-agnostic Warp message ID, and could be ported as-is to the protocol. To do so, we would [define](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/proto/p2p/p2p.proto) a new P2P `Message` type:
+
+```diff
+message Message {
+    ...
+    // App messages:
+    AppRequest app_request = 30;
+    AppResponse app_response = 31;
+    AppGossip app_gossip = 32;
+    AppError app_error = 34;
+    
++   // Warp signature messages:
++   MessageSignatureRequest = 35;
+}
+```
+
+Where `MessageSignatureRequest` is defined identically to Subnet EVM/Coreth:
+```
+message MessageSignature {
+    bytes message_id = 1;
+}
+```
+
+From there, the `MessageSignatureRequest` would be forwarded to the VM to retreive the signature.
+
+## Open Questions
+
+How should signature requests be forwarded to the VM? Here are a couple of candidates:
+- RPC Chain VM and Platform VM use the `Signer` [service](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/proto/warp/message.proto) to issue and response to signature requests. Subnet EVM and Coreth could be modified to use this service.
+- Subnet EVM and Coreth currently [implement](https://github.com/ava-labs/subnet-evm/blob/v0.6.6/peer/network.go) `AppHandler` to directly receive other p2p messages. A `SignatureHandler` interface could be provided for VMs to implement.
+
+In addition to generic Warp message signature request handling, Subnet EVM and Coreth also implement handling for Warp messages whose payload consists solely of a block hash. This is provides first class support for state verification via Warp in the VM. The proposed `MessageSignatureRequest` p2p message type does not distinguish between Warp message payload types, so it would be left to the VM to make that distinction.
+
+## Security Considerations
+
+## Backwards Compatibility
+This change is backwards compatible, though nodes that do not implement `MessageSignatureRequest` as described will drop incoming requests. 
+
+## Reference Implementation
+
+A full reference implementation has not been provided yet. It must be provided prior to this ACP being considered "Implementable".
+
+## Acknowledgements
+Thanks to @joshua-kim and @iansuvak for discussion and feedback on this ACP.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+From there, the `MessageSignatureRequest` would be forwarded to the VM to retreive the signature.
+
+## Open Questions
+
+How should signature requests be forwarded to the VM? Here are a couple of candidates:
+- RPC Chain VM and Platform VM use the `Signer` [service](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/proto/warp/message.proto) to issue and response to signature requests. Subnet EVM and Coreth could be modified to use this service.
+- Subnet EVM and Coreth currently [implement](https://github.com/ava-labs/subnet-evm/blob/v0.6.6/peer/network.go) `AppHandler` to directly receive other p2p messages. A `SignatureHandler` interface could be provided for VMs to implement.
+
+In addition to generic Warp message signature request handling, Subnet EVM and Coreth also implement handling for Warp messages whose payload consists solely of a block hash. This is provides first class support for state verification via Warp in the VM. The proposed `MessageSignatureRequest` p2p message type does not distinguish between Warp message payload types, so it would be left to the VM to make that distinction.
```
iansuvak, line 65:
```suggestion
In addition to generic Warp message signature request handling, Subnet EVM and Coreth also implement handling for Warp messages whose payload consists solely of a block hash. This provides first class support for state verification via Warp in the VM. The proposed `MessageSignatureRequest` p2p message type does not distinguish between Warp message payload types, so it would be left to the VM to make that distinction.
```

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,81 @@
...
+message Message {
+    ...
+    // App messages:
+    AppRequest app_request = 30;
+    AppResponse app_response = 31;
+    AppGossip app_gossip = 32;
+    AppError app_error = 34;
+    
++   // Warp signature messages:
++   MessageSignatureRequest = 35;
```
iansuvak, line 46: What would the downside be of expanding this to `SignatureRequest` and having it defined as below?

```
message SignatureRequest {
    uint32 type_id = 1
    bytes id = 2
}
```

Perhaps a better name for the second `id` field should be decided but this would then generalize to both message and hash signature requests. 

You mention below in `Open Questions` that interpreting the ID could be left to the VM but in that case we would have  a naming clash where current `AppRequest` based implementation distinguishes based on the message type and has both `MessageSignatureRequest` as well as `BlockSignatureRequest`.

[Warp docs](https://github.com/ava-labs/avalanchego/tree/576b3927c54b0e7ad91602572f4aa8f9fe24d768/vms/platformvm/warp/payload) specify that `Hash` payloads are chain specific, but we wouldn't lose this generality in the proto definition. I.e. if another chain used the `Hash` `typeID` of `0x 0x00000000` for something other than the block hash this message type would still be usable for it and would disambiguate from `AddressedCall` messages.

iansuvak (commented):
I'm very excited to see this moving! I have a high level question about specificity of the proposed type, but I might be misunderstanding something as well.

ACPs/118-warp-signature-request/README.md
iansuvak: Maybe `signable_id`?

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,81 @@
...
+}
+```
+
+Where `MessageSignatureRequest` is defined identically to Subnet EVM/Coreth:
+```
+message MessageSignature {
+    bytes message_id = 1;
```
joshua-kim, line 53: This field implies that all messages will be similar to warp in that we don't request the signature of the message, but rather that the signature is performed over something else that the message id is used to look up.

I think we should let the application specify this and just call this something general like `message` or `bytes` or `payload`.

joshua-kim, line 53: If we are introducing this to the p2p layer I think this is because we want ACP-75 to be able to use these signature requests?

In that case how will we disambiguate between a message sent by consensus vs a message sent by the VM (at the p2p layer we can get a message but not know currently if the message is meant for consensus, or if it's meant for the VM)? I think we need some identifier here to disambiguate. I think some `[]byte` or `varint` `handler_id` field could work here.

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,81 @@
...
+```diff
+message Message {
+    ...
+    // App messages:
+    AppRequest app_request = 30;
+    AppResponse app_response = 31;
+    AppGossip app_gossip = 32;
+    AppError app_error = 34;
+    
++   // Warp signature messages:
```
joshua-kim, line 45: I think we are going to need a corresponding response type

ACPs/118-warp-signature-request/README.md
joshua-kim, line 53: I think we are going to also need a way to protect against unrequested responses hitting us. The sender can use something like a requester-maintained `request_id` field so that we drop messages for a `request_id` we don't know about, similar to `AppRequest` messages.

We will also need a way to route this message between VMs, as the p2p layer handles networking for the entire node we will need to have some field like `chain_id` so that we can route messages to the correct VM.

iansuvak: So the idea here would be that we always send the whole payload that we are requesting signed? That's certainly less ambiguous but also potentially complicates the logic to decide whether it's something that you are willing to sign. And if a validator is willing to sign something it's likely already aware of it somehow so we can avoid the network and serde overhead.

If we intend to use these messages for consensus related signing in addition to warp we could use two levels of `type_id` fields to specify the request?

cam-schultz:
The original ACP PR proposed implementing `MessageSignatureRequests` at the P2P level. The PR has been updated to instead propose standardizing `MessageSignatureRequest` as an `AppRequest` embedded message type.

ACPs/118-warp-signature-request/README.md
cam-schultz: Correct me if I'm wrong, but routing is no longer a concern since that is handled at the `AppRequest` level.

cam-schultz: Revised the PR to propose standardizing the `AppRequest` message type.

ACPs/118-warp-signature-request/README.md
cam-schultz: I added a new `HashSignatureRequest` type that encodes a hash over arbitrary bytes. That would be a candidate for extending with a `type_id` field that's interpretable by VMs. I think `MessageSignatureRequest` should be tightly scoped to Warp message IDs specifically.

cam-schultz: Alternatively, we could generalize `HashSignatureRequest` to `SignatureRequest` that specifies arbitrary bytes, and leave it to the VM to define how to decode. Not sure how useful that would be over decoding the `AppRequest` payload directly.

iansuvak: I agree now that we should be tightly scoped to Warp Message IDs specifically. Now that I think about it again, my initial reservations were influenced by the current `subnet-evm` implementation and I was parsing `MessageSignatureRequest` and `BlockSignatureRequest` as disjoint use-cases but really `MessageSignatureRequest` is the general one and `BlockSignatureRequest` is just a special case. 

Given that I'm fine with completely dropping my proposal of `type_id` fields and I'm not even convinced that we need the `HashSignatureRequest` either.

michaelkaplan13:
Could you update the PR title to include [ACP-118], and description to be best practices track?

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,84 @@
...
+Title: Warp Signature Interface Standard
+Author(s): Cam Schultz (https://github.com/cam-schultz)
+Discussions-To: 
+Status: Proposed
+Track: Best Practices Track
+```
+
+## Abstract
+
+Proposes a standard Warp signature request AppRequest types so that Warp signatures may be requested in a VM-agnostic manner. To make this concrete, this standard type should be defined in AvalancheGo such that VMs can import it at the source code level. This will also simplify external signature aggregator implementations by allowing them to depend only on AvalancheGo for message construction, rather than individual VM codecs.
```
michaelkaplan13, line 12: Could you link the code definition of `AppRequest` here, and specify that the standard being proposed is for the format of the `request` bytes of an `AppRequest` that signify a request for a Warp signature?

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Motivation
+
+Warp message signatures consist of an aggregate BLS signature composed of the individual signatures of a subnet's validators. Individual signatures need to be retreivable by the party that wishes to construct an aggregate signature. At present, this is left to VMs to implement, as is the case with [Subnet EVM](https://github.com/ava-labs/subnet-evm/blob/v0.6.6/plugin/evm/message/signature_request.go#20) and [Coreth](https://github.com/ava-labs/coreth/blob/v0.13.6-rc.0/plugin/evm/message/signature_request.go#L20)
+
+This creates friction in applications that are intended to operate across many VMs (or distinct implementations of the same VM). As an example, the reference Warp message relayer implementation, [awm-relayer](https://github.com/ava-labs/awm-relayer), fetches individual signatures from validators and aggregates them before sending the Warp message to its destination chain for verification. However, Subnet EVM and Coreth have distinct codecs, requiring the relayer to [switch](https://github.com/ava-labs/awm-relayer/blob/v1.4.0-rc.0/relayer/application_relayer.go#L372) according to the target codebase.
```
michaelkaplan13, line 18:
```suggestion
This creates friction in applications that are intended to operate across many VMs (or distinct implementations of the same VM). As an example, the reference Warp message relayer implementation, [awm-relayer](https://github.com/ava-labs/awm-relayer), fetches individual signatures from validators and aggregates them before sending the Warp message to its destination chain for verification. However, `subnet-evm` and `coreth` have distinct codecs, requiring the relayer to [switch](https://github.com/ava-labs/awm-relayer/blob/v1.4.0-rc.0/relayer/application_relayer.go#L372) according to the target codebase.
```

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Another example is ACP-75, which aims to implement acceptance proofs using Warp. The signature aggregation mechanism is not [specified](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/75-acceptance-proofs/README.md#signature-aggregation), which is a blocker for that ACP to be marked implementable.
+
+Standardizing the Warp Signature Request format by implementing it as an AppRequest type in AvlanahceGo would simplify the implementation of ACP-75, and streamline signature aggregation for out-of-protocol services such as Warp message relayers.
```
michaelkaplan13, line 22:
```suggestion
Standardizing the Warp Signature Request interface by defining it as a format for `AppRequest` message payloads in AvalanahceGo would simplify the implementation of ACP-75, and streamline signature aggregation for out-of-protocol services such as Warp message relayers.
```

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,90 @@
...
+type SignatureResponse struct {
+	Signature [bls.SignatureLen]byte
+}
+```
+
+`MessageID` represents the VM-agnostic Warp message ID, while `BlockID` represents a block hash that nodes are willing to attest to via Warp signature. and could be ported as-is to AvalancheGo. `BlockID` can instead be generalized to represent a 32-byte hash over some data. Interpreting that hash is left to VMs. For example, `Hash` could represent a transaction hash that a node only signs on receiving a `HashSignatureRequest`. 
+
+## Specificaiton
+
+We take the types currently defined in Subnet EVM and Coreth, modify them to not be EVM specific, and implement them as `AppRequest` message types will be defined as Protobuf specs. Corresponding `AppRequest` handlers will need to be implemented.
```
michaelkaplan13, line 51: This first sentence seems like a run-on, could you reword?

Also I think it should clarify that AppRequest handlers are left for the VMs to implement in order to adhere to this interface.

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Subnet EVM and Coreth also support attesting to block hashes via Warp, by serving signature requests made using the following `AppRequest` type:
+
+```
+type BlockSignatureRequest struct {
+	BlockID ids.ID
+}
+```
+
+`SignatureRequest` could achieve this by specifying an unsigned Warp message with the `BlockID` as the payload, and serializing that message into the `data` field. `justification` could optionally be used to provide additional context, such as a lookback window.
```
michaelkaplan13, line 77:
```suggestion
`SignatureRequest` could achieve this by specifying an unsigned Warp message with the `BlockID` as the payload, and serializing that message into the `data` field. `justification` could optionally be used to provide additional context, such as a the block height of the given block ID.
```

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+```
+type MessageSignatureRequest struct {
+	MessageID ids.ID
+}
+```
+
+For all Warp messages that the node has seen, (i.e. on-chain message sent through the [Warp Precompile](https://github.com/ava-labs/subnet-evm/tree/v0.6.7/precompile/contracts/warp) and [off-chain](https://github.com/ava-labs/subnet-evm/blob/v0.6.7/plugin/evm/config.go#L226) Warp messages) there exists a mapping of Warp message ID to the full unsigned message bytes. Subnet EVM and Coreth nodes are therefore able to retreive and sign the full Warp message given its message ID.
+
+`SignatureRequest` could be used for this case by specifying the Warp message ID as the `justification` field, and omitting the `data` field.
```
michaelkaplan13, line 65: I guess it's up to the implementations either way, but I find it a bit confusing that the `data` field is considered optional since it's otherwise thought of as the data to be signed. 

I was picturing a case where the raw data to be signed is not included in the request as a separate interface, not part of this standard.

cam-schultz: Reworded the specification such that `data` is not optional, and removed the "request a signature by message ID" use case.

michaelkaplan13 (approved):

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,88 @@
...
+`SignatureRequest` can achieve this by specifying an unsigned Warp message with the `BlockID` as the payload, and serializing that message into `data`. `justification` may optionally be used to provide additional context, such as a the block height of the given block ID.
+
+### Confirm that an event did not occur
+
+With [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets), Subnets will have the ability to manage their own validator sets. The Warp message payload contained in a `RegisterSubnetValidatorTx` includes an `expiry`, after which the specified validation ID becomes invalid. The Subnet needs to know that this validation ID is expired so that it can keep its locally tracked validator set in sync with the P-Chain. We also assume that the P-Chain will not persist expired or invalid validation IDs. 
+
+We can use `SignatureRequest` to construct a Warp message attesting that the validation ID expired. We do so by serializing an unsigned Warp message containing the validation ID into `data`, and providing additional metadata in `justification` for the P-Chain to reconstruct the expired validation ID.
+
+## Security Considerations
+TODO
```
aaronbuchwald, line 74: Should complete this section before merging even in draft imo

ACPs/118-warp-signature-request/README.md
```diff
@@ -0,0 +1,88 @@
...
+## Specification
+
+We propose the following types, implemented as Protobuf types that may be decoded from the `AppRequest`/`AppResponse` `app_bytes` field. By way of example, this approach is currently used to [implement](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/proto/sdk/sdk.proto#7) and [parse](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/network/p2p/gossip/message.go#22) gossip `AppRequest` types.
+
+- `SignatureRequest` includes two fields. `data` specifies the payload that the returned signature should correspond to, namely a serialized unsigned Warp message. `justification` is an optional field that specifies arbitrary data that the requested node may use to decide whether or not it is willing to sign `data`.
+
+    ```protobuf
+    message SignatureRequest struct {
+        bytes data = 1;
+        bytes justification = 2;
```
aaronbuchwald, line 33: How will the `justification` field be used?

If the justification is going to provide a hint like a blockID/messageID, should this data be typed and standardized as well?

aaronbuchwald: Could we also be more specific about what it means for this to be an optional field? I think it's better to be more prescriptive for how VMs should use this.

Ie. if the justification includes a blockID, does the VM only serve the request if the justification is provided or is it responsible for parsing the message and figuring out what should have been there if it's not provided?

If we're very explicit upfront, then I think this will make it simpler for VM implementers and integrators that can use a uniform standard rather than needing to adjust to any ambiguity that is left up to the VM to decide.

ACPs/118-warp-signature-request/README.md
michaelkaplan13: Agreed. Think the only consideration here is that VMs implementing the proposed standard interface need to consider how to verify what data they should be willing to sign, and if the implementation is DOSable. The implementation is left to the VMs to provide themselves though.

ACPs/118-warp-signature-request/README.md
cam-schultz: > should this data be typed and standardized as well?

It's completely up to the VM to interpret the `justification` in the context of the Warp payload format contained in `data`. Both `data` and `justification` could be parsed into concrete types by the VM, but I don't think we can specify a type for `justification` in `SignatureRequest` without losing generality.

> Could we also be more specific about what it means for this to be an optional field?

Absolutely. I think an illustrative example showing how `data` and `justification` should resolve to coupled types on the VM side could be useful. I'll put that together.

ACPs/118-warp-signature-request/README.md
cam-schultz: Added a list of security considerations that VM implementers should take into account.

ACPs/118-warp-signature-request/README.md
cam-schultz: I expanded the description of the fields in the Specification to make it clear that it is the VM's responsibility to define the requirements for `data` and `justification`. I also clarified the `Confirm that an event did not occur` use case to make the example more concrete.

PR #129, opened 2024-07-23
ACP-118: Specify SignatureRequest handler ID

cam-schultz:
Updates the ACP-118 specification to include the canonical handler ID that should be used by all VMs to register `SignatureRequest` handler implementations.

Specifies the canonical handler ID as `1`, incrementing from the only other handler ID [registration](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/network/p2p/network.go#L173), which uses a [handler ID](https://github.com/ava-labs/avalanchego/blob/v1.11.10-status-removal/vms/platformvm/network/network.go#L24) of `0`.

### DIFF
--- a/ACPs/118-warp-signature-request/README.md
+++ b/ACPs/118-warp-signature-request/README.md
@@ -42,7 +42,9 @@ We propose the following types, implemented as Protobuf types that may be decode
     }
     ```
 
-For each of these types, VMs must implement corresponding `AppRequest` and `AppResponse` handlers to adhere to the interface.
+### Handlers
+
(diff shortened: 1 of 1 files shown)


iansuvak (approved):

michaelkaplan13 (commented):
Want to link the AvalancheGo PR with the reference implementation and mark this ACP as "implementable"?

cam-schultz:
> Want to link the AvalancheGo PR with the reference implementation and mark this ACP as "implementable"?

My slight preference would be to link to the merged source code, rather than a pull request. If this gets merged before the AvalancheGo PR, then we can follow up with another PR to mark the ACP as implementable.

tsachiherman (approved):
looks good to me, but would be happy to have others review this as well.

PR #135, opened 2024-07-29
[ACP-77] Updates

dhrubabasu:
* Readability Improvements
* Clarifications on `CreateSubnetTx` flow to Permissionless Subnet
* Specify warp messages proving subnet validator registration on P-Chain
* Wording nits


ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
-|    expiry :   uint64 |   8 bytes |
-+-----------+----------+-----------+
-| signature : [64]byte |  64 bytes |
-+-----------+----------+-----------+
-                       | 144 bytes |
-                       +-----------+
+To provide maximal flexibility for Permissionless Subnets, the BLS multisignature approach in Avalanche Warp Messaging (AWM) is re-used to approve modifications to the Subnet's validator set. Using the `(chainID, address)` pair defined in the `ConvertSubnetTx`, a Warp Message with an [`AddressedCall`](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/warp/payload#addressedcall) payload can be constructed.
+
+To validate an `AddressedCall` payload in Avalanche Warp Messaging, the `(chainID, address)` pair is used to lookup the validators of `chainID` and verify that the BLS multi-sig includes a quorum (set to 67%) of the Subnet's validator set. The P-Chain will use Warp Messages with `AddressedCall` payloads to support modifications of the Subnet's validator set using this `(chainID, address)` pair (using the `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` defined later in the specification).
```
michaelkaplan13, line 57: Should it be specified here that the `sourceAddress` of the `AddressedCall` payload is expected to have 0 length?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
 
-- Use the number of tokens the user has staked into a smart contract on the Subnet to determine the weight of their validator
-- Require the user to submit an on-chain transaction with their validator information
-- Generate the Warp message
+A `RecoverSubnetTx` can be used in this situation to instantiate the Subnet's validator set using the `Owner` key defined in `CreateSubnetTx`. In all other situations, the `Owner` key is powerless after a `ConvertSubnetTx` is issued.
+
+Each Subnet Validator will be instantiated with a Balance of `(sum($AVAX inputs) - sum($AVAX outputs) - TxFee) / len(Validators)` rounded down. For a `RecoverSubnetTx` to be valid, `(sum($AVAX inputs) - sum($AVAX outputs) - TxFee) / len(Validators)` must be >= the greater of 5 $AVAX or two weeks of the current fee. This prevents Subnet Validators from being added with too low of an initial balance where they become immediately delinquent based on the continous fee mechanism defined below. A Subnet Validator can leave at any time before the initial $AVAX is consumed and claim the remaining balance to the `ChangeOwner` defined in the transaction.
```
michaelkaplan13, line 86: Super nit: Do we want to capitalize Validator? I don't think I've seen it capitalized elsewhere

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
+
+    `validationID` is defined as the SHA256 hash of the `Payload` of the `AddressedCall`. This SHA256 hash will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` is required to be no longer than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` with `expiry` more than 48 hours in the past can be flushed from the P-Chain's state.
+- `signature` is the raw bytes of the Ed25519 signature over the concatenated bytes of `[subnetID]+[nodeID]+[blsPublicKey]+[weight]+[balance]+[expiry]`. This signature must correspond to the Ed25519 public key that is used for the `nodeID`. This approach prevents NodeIDs from being unwillingly added to Subnets. `balance` is the minimum initial $nAVAX balance that must be attached to the validator serialized as a uint64. A Subnet may choose to validate that the `signature` field is well-formed but it is not required.
```
michaelkaplan13, line 175:
```suggestion
- `signature` is the raw bytes of the Ed25519 signature over the concatenated bytes of `[subnetID]+[nodeID]+[blsPublicKey]+[weight]+[balance]+[expiry]`. This signature must correspond to the Ed25519 public key that is used for the `nodeID`. This approach prevents NodeIDs from being unwillingly added to Subnets. `balance` is the minimum initial $nAVAX balance that must be attached to the validator serialized as a uint64. A Subnet may choose to validate that the `signature` field is well-formed, but it is not required.
```

Should we move this bullet point above the `validationID` description? Seems a little odd in the middle of the field descriptions.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
+Once this transaction is accepted, `AddSubnetValidatorTx` and `RemoveSubnetValidatorTx` are disabled on the Subnet. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward.
 
-Subnets are responsible for defining the procedure on how to retrieve the above information from prospective validators.
+### Recovering a Subnet
 
-An EVM Subnet may choose to implement this step like so:
+In the event that a Subnet has no validators, a valid BLS multi-signature cannot be produced. This situation can arise immediately after a Subnet is created or if all the Subnet's validators were removed from the validator set via `SetSubnetValidatorWeightTx` with `Weight = 0`.
```
michaelkaplan13, line 82:
```suggestion
In the event that a Subnet has no validators, a valid BLS multi-signature cannot be produced. This is the case immediately after a Subnet is created, and can arise later if all the Subnet's validators are removed from the validator set.
```

Wording nit. On the second part, validators can be removed via `SetSubnetValidatorWeightTx` or `ExitValidatorSetTx` right? Don't think it's necessary to specify both here, but mostly don't think we should include one without the other.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
 
-- Use the number of tokens the user has staked into a smart contract on the Subnet to determine the weight of their validator
-- Require the user to submit an on-chain transaction with their validator information
-- Generate the Warp message
+A `RecoverSubnetTx` can be used in this situation to instantiate the Subnet's validator set using the `Owner` key defined in `CreateSubnetTx`. In all other situations, the `Owner` key is powerless after a `ConvertSubnetTx` is issued.
```
cam-schultz, line 84: Is `SubnetAuth` in `RecoverSubnetTx` and `ConvertSubnetTx` the same as `Owner`?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
 
-After the `RegisterSubnetValidatorTx` is accepted on the P-Chain, the Subnet Validator is added to the Subnet's validator set. The P-Chain will then send a Warp message to the Subnet notifying of the validator set addition.
+The `Message` field must be an `AddressedCall` with the payload:
+
+```text
++-----------+----------+-----------+
+|   codecID :   uint16 |   2 bytes |
++-----------+----------+-----------+
+|    typeID :   uint32 |   4 bytes |
++-----------+----------+-----------+
```
cam-schultz, line 154: These two fields are not represented in the type spec comment above.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -132,37 +209,55 @@ Since there are no `EndTime`s enforced by the P-Chain, Subnets must use this tra
...
+- `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
+- `typeID` is the payload type identifier and is `0x00000001` for this transaction
+- `validationID` is the SHA256 of the `Payload` of the `AddressedCall` in the `RegisterSubnetValidatorTx` adding the validator to the Subnet's validator set
```
cam-schultz, line 240: Are the `codecID` and `typeID` included in the `SHA256`?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -177,41 +272,60 @@ type ExitValidatorSetTx struct {
...
 
-To participate in this new registration flow, Subnets must set the `(blockchainID, address)` pair by using `SetSubnetValidatorManagerTx`. This pair will be used to validate the `AddressedCall` in `SetSubnetValidatorWeightTx` and `RegisterSubnetValidatorTx`.
+To track whether a Subnet Validator addition/modification/removal occured on the P-Chain, Subnet Validators must be willing to sign an `AddressedCall` attesting to the validator set change. Since all Subnet Validators sync the P-Chain, only the Subnet Validators need to sign the `AddressedCall`, not all Primary Network Validators.
 
-When a Subnet is created with `CreateSubnetTx`, a validator manager is not specified. To solve this problem, the first `SetSubnetValidatorManagerTx` will expect the Warp `AddressedCall` to have the `sourceChainID` be set to the `SubnetID` and the `sourceAddress` to be zero'd out. Once this transaction is accepted, `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` are enabled for use on the Subnet.
+Two `AddressedCall`s are defined with the below payloads. For each of them, the `sourceChainID` must be set to the P-Chain ID and the `sourceAddress` must be set to an empty byte array.
```
cam-schultz, line 279: Consider clarifying that `sourceAddress` is in the `AddressedCall`, and `sourceChainID` is in the top-level Warp message.

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald: should be lowercase

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald: `SubnetAuth` is necessary for any transactions that require signing with an Owner Key. Many of the existing p-chain transactions already implement this, i.e. [CreateChainTx](https://github.com/ava-labs/avalanchego/blob/a0f7e422bb8497186ebff434f12c4b957cec8d49/vms/platformvm/txs/create_chain_tx.go#L50). Here's an [example](https://github.com/ava-labs/avalanchego/blob/a0f7e422bb8497186ebff434f12c4b957cec8d49/wallet/chain/p/signer/visitor.go#L96) made with the p chain wallet.

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald: https://github.com/dhrubabasu/ACPs/pull/1#discussion_r1696100009

ACPs/77-reinventing-subnets/README.md
cam-schultz: Thanks for the context. I commented because it wasn't clear to me how `Owner` maps to the code. Any docs that describe the relationship between `Owner` and `SubnetAuth` would be helpful.

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald: IMO it should be reflected in the type spec to avoid confusion, otherwise we will probably get more questions about this

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
 
-#### Step 1: Retrieve a BLS multisig from the Subnet
+To create a Subnet, a `CreateSubnetTx` must be issued on the P-Chain. This transaction includes an `Owner` field which defines the key that must be used to authorize any validator set additions (`AddSubnetValidatorTx`) or removals (`RemoveSubnetValidatorTx`).
 
-To provide increased optionality over Subnet Validator requirements, the P-Chain will only require an Avalanche Warp Message using the `AddressedCall` format with the below payload to join the Subnet Validator set. A `typeID` for distinguishing between the different payload types on the P-Chain must be included prior to this ACP being considered "Implementable".
+To be a Permissionless Subnet, this `Owner` key must no longer have the ability to modify the Subnet's validator set. A `ConvertSubnetTx` must first be issued to explicitly convert a Subnet from Permissioned to Permissionless. This transaction will set the `(chainID, address)` pair that will manage the Subnet going forward. After `ConvertSubnetTx` is issued, the `Owner` from the `CreateSubnetTx` that created the Subnet will no longer have the ability to modify the Subnet's validator set.
```
meaghanfitzgerald, line 53: Would a `[CreateChainTx`](https://github.com/ava-labs/avalanchego/blob/a0f7e422bb8497186ebff434f12c4b957cec8d49/vms/platformvm/txs/create_chain_tx.go#L36) need to be issued _before_ `ConvertSubnetTx`?

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald: definitely need better docs on this. afaict, the [`verifySubnetAuthorization`](https://github.com/ava-labs/avalanchego/blob/a0f7e422bb8497186ebff434f12c4b957cec8d49/vms/platformvm/txs/executor/subnet_tx_verification.go#L53) uses [`getSubnetOwner`](https://github.com/ava-labs/avalanchego/blob/a0f7e422bb8497186ebff434f12c4b957cec8d49/vms/platformvm/state/diff.go#L279) to get the owner from state. This was set by [`setSubnetOwner`](https://github.com/ava-labs/avalanchego/blob/a0f7e422bb8497186ebff434f12c4b957cec8d49/vms/platformvm/txs/executor/standard_tx_executor.go#L107) when `CreateSubnetTx` was executed.

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald, line 86:
```suggestion
Each Subnet validator will be instantiated with a Balance of `(sum($AVAX inputs) - sum($AVAX outputs) - TxFee) / len(Validators)` rounded down. For a `RecoverSubnetTx` to be valid, `(sum($AVAX inputs) - sum($AVAX outputs) - TxFee) / len(Validators)`(the initial Balance per validator) must be >= the greater of 5 $AVAX or two weeks of the current fee. This prevents Subnet validators from being added with too low of an initial balance where they become immediately delinquent based on the continuous fee mechanism defined below. A Subnet validator can leave at any time before the initial $AVAX is consumed and claim the remaining balance to the `ChangeOwner` defined in the transaction.
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -177,41 +272,60 @@ type ExitValidatorSetTx struct {
...
+                          +----------+
 ```
 
-A `typeID` for distinguishing between the different payload types on the P-Chain must be included prior to this ACP being considered "Implementable".
+- `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
+- `typeID` is the payload type identifier and is `0x00000002` for this message
+- `validationID` is the SHA256 of the `Payload` of the `AddressedCall` in the `RegisterSubnetValidatorTx` adding the validator to the Subnet's validator set
+- `valid` indicates whether or not the `validationID` corresponds to a valid `AddressedCall` payload
+
+#### SetSubnetValidatorWeightMessage
```
Nuttymoon, line 304: To follow the same phrasing as `SubnetValidatorRegistrationMessage`:
```suggestion
#### SubnetValidatorWeightUpdateMessage
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -177,41 +272,60 @@ type ExitValidatorSetTx struct {
...
++--------------+----------+----------+
+                          | 39 bytes |
+                          +----------+
 ```
 
-A `typeID` for distinguishing between the different payload types on the P-Chain must be included prior to this ACP being considered "Implementable".
+- `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
+- `typeID` is the payload type identifier and is `0x00000002` for this message
+- `validationID` is the SHA256 of the `Payload` of the `AddressedCall` in the `RegisterSubnetValidatorTx` adding the validator to the Subnet's validator set
+- `valid` indicates whether or not the `validationID` corresponds to a valid `AddressedCall` payload
```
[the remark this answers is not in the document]
Nuttymoon: IMO the name of the field is confusing as well. Maybe `registered` would make more sense?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -177,41 +272,60 @@ type ExitValidatorSetTx struct {
...
-    // Warp message should include:
-    //   - SubnetID
-    //   - ChainID (where the validator manager lives)
-    //   - Address (address of the validator manager)
-    //   - BLS multisig over the above payload
-    Message warp.Message `json:"message"`
-}
+The method of requesting is left unspecified as it can be more generic. A viable option for supporting this functionality is laid out in [ACP-118](../118-warp-signature-request/README.md) with the `SignatureRequest` message.
+
+#### SubnetValidatorRegistrationMessage
```
Nuttymoon, line 283: I am wondering if having 2 payloads is really needed.  
Can't we always use the weight update payload below? Even to check for successful registration with `nonce` = `0`.

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
Nuttymoon: Also shouldn't it be possible to change the Subnet Manager after the `ConvertSubnetTx`? e.g. to change the address of the manager?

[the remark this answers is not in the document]
dhrubabasu: How does this sound to you?

```suggestion
A `RecoverSubnetTx` can be used in this situation to instantiate the Subnet's validator set using the Subnet's current `Owner`. In all other situations, the `Owner` key is powerless after a `ConvertSubnetTx` is issued. The only exception is for `Owner` rotation via the `TransferSubnetOwnershipTx`.
```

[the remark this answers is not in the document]
dhrubabasu: > Also shouldn't it be possible to change the Subnet Manager after the ConvertSubnetTx? e.g. to change the address of the manager?

Rotating the manager address has a ton of issues that need to be worked out (e.g. how do you migrate the validator set and associated stake from the original address to the new address?). The original draft of ACP-77 had a manager rotation but it was removed because of that additional complexity it added to an already large ACP.

ACPs/77-reinventing-subnets/README.md
dhrubabasu: YEs

ACPs/77-reinventing-subnets/README.md
dhrubabasu: By virtue of being an `AddressedCall` payload, there is a `codecID` and `typeID` (Ref: https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/warp/payload#addressedcall)

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
michaelkaplan13: Yeah it's definitely kinda odd because there are more than two possible states but we're using a `bool`.

The 3 states are:
1. Currently active validation period.
2. Not a currently active validation period, but could possibly become one in the future (not yet passed its registration expiry time)
3. Not a currently active validation period, and will never be an active validation period (due to the expire being passed)

No message should be willing to be signed for validation IDs in case 2, so there isn't a need represent it at all.

I agree `registered` is more clear though.

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
Nuttymoon: Sounds good!

It makes sense to remove the manager's address rotation. I agree with the added complexity.

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
felipemadero: does the chain referenced by chainID needs to belong to the subnet to be transformed? (it kind of seems so from the description `the (chainID,address) pair is used to lookup the validators of chainID`)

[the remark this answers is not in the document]
felipemadero: how is P-Chain using the chainID in the (chainID, address) pair when it receives a RegisterSubnetValidatorTx or a SetSubnetValidatorWeightTx, if it is the case that the chainID may not belong to the subnet to be transformed?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
+    // Note: We do not enforce that the BLS key is unique across all validators.
+    //       This means that validators can share a key if they so choose.
+    //       However, a NodeID + Subnet does uniquely map to a BLS key
+    Signer signer.Signer `json:"signer"`
+    // Leftover $AVAX from the [Balance] will be issued to this
+    // owner once it is removed from the validator set.
+    ChangeOwner fx.Owner `json:"changeOwner"`
+}
+
+type RecoverSubnetTx struct {
```
felipemadero, line 106: maybe Recover is not the best name for the case of first using it, after a convert.  maybe the starting set of validators can be just set on Convert itself?

ACPs/77-reinventing-subnets/README.md
dhrubabasu: The `sourceAddress` is not expected to have 0 length for `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx`. It is expected to have 0 length for the message from P-Chain -> Subnet which is addressed in the `Proof of Subnet Validator Set Change` section

ACPs/77-reinventing-subnets/README.md
dhrubabasu: I've been using uppercase here to follow "Primary Network Validator"

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Removed `signature` from warp message to use `blsPublicKey` directly.

ACPs/77-reinventing-subnets/README.md
dhrubabasu: > Would a CreateChainTx need to be issued before ConvertSubnetTx?

Yes

dhrubabasu: > Also wondering if a Subnet can be permissionless from its creation or if CreateChainTx will eventually accept a Subnet Manager that is not Owner from the get go

Currently no. It starts as Permissioned then can be migrated to Permissionless. ACP-77 will probably stop here in terms of Subnet creation flow since it's already fairly involved. The UX streamlining you are describing would be better suited in a follow up ACP

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
+    // Weight of this validator used when sampling
+    Weight uint64 `json:"weight"`
+    // [Signer] is the BLS key for this validator.
+    // Note: We do not enforce that the BLS key is unique across all validators.
+    //       This means that validators can share a key if they so choose.
+    //       However, a NodeID + Subnet does uniquely map to a BLS key
+    Signer signer.Signer `json:"signer"`
+    // Leftover $AVAX from the [Balance] will be issued to this
+    // owner once it is removed from the validator set.
+    ChangeOwner fx.Owner `json:"changeOwner"`
```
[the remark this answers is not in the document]
dhrubabasu: Any suggestions? I picked ChangeOwner to match existing nomenclature 🤷🏾‍♂️

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Existing PoA Subnets should be able to keep their validator set through the ConvertSubnetTx so I didn't include the starting set on ConvertSubnetTx.

Recover signifies that this is a transaction that is only used as last-resort. Open to other suggestions

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
dhrubabasu: The chain can be any chain. RegisterSubnetValidatorTx and SetSubnetValidatorWeightTx is only valid if the (chainID, address) is equivalent to what is stored on the P-Chain for the Subnet manager

ACPs/77-reinventing-subnets/README.md
dhrubabasu: I don't think it's strictly needed. However, I think it's cleaner to have two separate ones to correspond to the two different warp messages used in `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx`.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
-+-----------+----------+-----------+
-|    weight :   uint64 |   8 bytes |
-+-----------+----------+-----------+
-|    expiry :   uint64 |   8 bytes |
-+-----------+----------+-----------+
-| signature : [64]byte |  64 bytes |
-+-----------+----------+-----------+
-                       | 144 bytes |
-                       +-----------+
+To provide maximal flexibility for Permissionless Subnets, the BLS multisignature approach in Avalanche Warp Messaging (AWM) is re-used to approve modifications to the Subnet's validator set. Using the `(chainID, address)` pair defined in the `ConvertSubnetTx`, a Warp Message with an [`AddressedCall`](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/warp/payload#addressedcall) payload can be constructed.
```
minghinmatthewlam, line 55: just a clarification on wording is "BLS multisignature" the same as "BLS aggregate signature"? I haven't heard of it referred to as a multisig before.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -132,37 +209,55 @@ Since there are no `EndTime`s enforced by the P-Chain, Subnets must use this tra
...
+This transaction can be issued on the P-Chain without explicit consent of the Subnet. It is expected that validators should be removed from the Subnet's validator set through a `SetSubnetValidatorWeightTx` with weight `0` by initiating Subnet Validator removal on the Subnet. However, the ability to exit a Subnet Validator set is critical for censorship-resistance and/or failed Subnets. If a validator ever wishes to stop participating in Subnet consensus, they will be able to do so through this transaction. This is enforced on the P-Chain to prevent Subnets from locking Validators into participating in consensus indefinitely. Note that this does not modify a Subnet's total staking weight, this transaction moves the Validator from Active to Inactive.
```
minghinmatthewlam, line 258: Wanted to ask for clarification on `Note that this does not modify a Subnet's total staking weight, this transaction moves the Validator from Active to Inactive.`

because above it says `Remaining $AVAX in the Subnet Validator's `Balance` will be issued to the `ChangeOwner`, so can't the left over avax be removed from the Subnet's total staking weight?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -231,15 +345,15 @@ type IncreaseBalanceTx struct {
...
 
 This new relationship between the P-Chain and Subnets provides a dynamic where Subnets can use the P-Chain as an impartial judge to modify parameters (in addition to its existing role of helping to validate incoming Avalanche Warp Messages). If a Validator is misbehaving, the Subnet Validators can collectively generate a BLS multisig to reduce the voting weight of a misbehaving validator. This operation is fully secured by the Avalanche Primary Network (225M $AVAX or $8.325B at the time of writing).
 
-Follow-up ACPs could extend the P-Chain <> Subnet relationship to include parametrization of the 67% threshold in `SetSubnetValidatorWeightTx` to enable Subnets to choose a different threshold based on their security model (e.g. a simple majority of 51%).
+Follow-up ACPs could extend the P-Chain <> Subnet relationship to include parametrization of the 67% threshold to enable Subnets to choose a different threshold based on their security model (e.g. a simple majority of 51%).
```
minghinmatthewlam, line 356: just curious, but could the idea be to include the Warp quorum in either CreateSubnetTx or ConvertSubnetTx?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -299,9 +299,9 @@ The method of requesting is left unspecified as it can be more generic. A viable
 - `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
 - `typeID` is the payload type identifier and is `0x00000002` for this message
 - `validationID` is the SHA256 of the `Payload` of the `AddressedCall` in the `RegisterSubnetValidatorTx` adding the validator to the Subnet's validator set
-- `valid` indicates whether or not the `validationID` corresponds to a valid `AddressedCall` payload
+- `registered` indicates whether or not the `validationID` corresponds to a valid `AddressedCall` payload. If true, `validationID` corresponds to an active validator. If false, `validationID` does not correspond to an active validator, and never will as the `expiry` in the `AddressedCall` payload is in the past.
```
Nuttymoon, line 302: What if the `validationID` is not active yet, but the `expiry` is not in the past?

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
felipemadero: Yes surely. I can suggest something on the lines of ValidatorLeftOverBalanceOwner

ACPs/77-reinventing-subnets/README.md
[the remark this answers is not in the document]
felipemadero: Ok

ACPs/77-reinventing-subnets/README.md
felipemadero: It seems the case that is going to be used for the typical flow, not as last resort. Maybe SetInitialSOV?

ACPs/77-reinventing-subnets/README.md
minghinmatthewlam: nvm I mixed up weight and balanace. IIUC, "total staking weight" is referring to the Subnet's defined staking calculations, and "balance" that is moved to ChangeOwner is referred to the P-Chain AVAX balance.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,154 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
-- Use the number of tokens the user has staked into a smart contract on the Subnet to determine the weight of their validator
-- Require the user to submit an on-chain transaction with their validator information
-- Generate the Warp message
+A `RecoverSubnetTx` can be used in this situation to instantiate the Subnet's validator set using the Subnet's current `Owner`. In all other situations, the `Owner` key is powerless after a `ConvertSubnetTx` is issued. The only exception is for `Owner` rotation via the `TransferSubnetOwnershipTx`.
+
+Each Subnet Validator will be instantiated with a Balance of `(sum($AVAX inputs) - sum($AVAX outputs) - TxFee) / len(Validators)` rounded down. For a `RecoverSubnetTx` to be valid, `(sum($AVAX inputs) - sum($AVAX outputs) - TxFee) / len(Validators)` must be >= the greater of 5 $AVAX or two weeks of the current fee. This prevents Subnet Validators from being added with too low of an initial balance where they become immediately delinquent based on the continuous fee mechanism defined below. A Subnet Validator can leave at any time before the initial $AVAX is consumed and claim the remaining balance to the `ChangeOwner` defined in the transaction.
```
meaghanfitzgerald, line 86: where is `Balance` input in `RecoverSubnetTx`?

ACPs/77-reinventing-subnets/README.md
michaelkaplan13, line 258: Nit: the capitalization of "Validator" is inconsistent here.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: In this case, the P-Chain nodes should refuse to the sign the message.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: The `codecID` and `typeID` for the `AddressedCall` payload of the Warp message are different than the `codecID` and `typeID` for the payload within the `AddressedCall` body though, no?

michaelkaplan13 (commented):
I think we should add to `Security Considerations` to explain that `NodeID`s can be added to a Subnet's validator set involuntarily, but this does not impact any stake/rewards for that node. If the node wanted to later honestly join that subnet as a validator, it would just have to rotate it's key to get a new node ID, which is an explicit tradeoff for not needing to provide unique signatures using the node's key for each validation period. This could be thought of as a type of "griefing" attack that has non-zero cost (need to provide balance on the P-chain), but zero long term effect (the griefed node ID just needs to rotate its key if it wants to join the subnet as a validator later, and isn't otherwise negatively impacted).

ACPs/77-reinventing-subnets/README.md
dhrubabasu: The proper term is multisignature: https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Could be. Left it to a future ACP to explore this since the security considerations need to be weighed very carefully

ACPs/77-reinventing-subnets/README.md
dhrubabasu: added a note

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Yes, that's correct. The `codecID` and `typeID` here are for the payload within the `AddressedCall`.

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Good catch, added!

ACPs/77-reinventing-subnets/README.md
```diff
@@ -46,77 +46,160 @@ For standalone networks like the Avalanche Primary Network, this is done by conn
...
 
-For a `RegisterSubnetValidatorTx` to be valid, `Balance` must be >= the greater of 5 $AVAX or two weeks of the current fee. This prevents Subnet Validators from being added with too low of a `Balance` where they become immediately delinquent based on the continuous fee mechanism defined below. A Subnet Validator can leave at any time before the initial $AVAX is consumed and claim the remaining `Balance` to the `ChangeOwner` defined in the transaction.
+    `validationID` is defined as the SHA256 hash of the `Payload` of the `AddressedCall`. This SHA256 hash will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` is required to be no longer than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` with `expiry` more than 48 hours in the past can be flushed from the P-Chain's state.
```
geoff-vball, line 179: Why do we wait 48 hours past the `expiry` to flush the `validationID`?

dhrubabasu: for replay protection. If we flush it 24 hours after, somebody can reissue the same message corresponding to the `validationID` and it would be valid since `expiry` is 24 hours in the future.

geoff-vball: I'm not really understanding. If we flush it 24 hours after the expiry, the expiry is 24 hours in the past, not in the future.

dhrubabasu: Ah I understand the confusion- this was poorly worded on my part.
```suggestion
    `validationID` is defined as the SHA256 hash of the `Payload` of the `AddressedCall`. This SHA256 hash will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` is required to be no longer than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` with an `expiry` in the past can be flushed from the P-Chain's state.
```

PR #146, opened 2024-09-10
[ACP-77] Remove RecoverSubnetTx. Add SubnetConversionMessage

michaelkaplan13:
The existing `recoverSubnetTx` is meant to allow the "owner" of the subnet on the P-Chain to specify the initial PAYG validator set. After the `convertSubnetTx`, the subnet owner would be powerless except in the case when the subnet removes all of its validators, in which case, they could send another `recoverSubnetTx`.

However, this design as previously unforeseen complexities:
- It's possible to have PAYG validators on a non-converted Subnet.
- These initial validators have a specified endTime, but their balance may run out prior to their end time.

Instead, we can remove the `RecoverSubnetTx` all together in favor of enforcing an invariant that a Subnet always has at least one PAYG validator after a `ConvertSubnetTx` has been issued for it.

This also simplifies the initialization flow for validator manager contracts like those defined in ACP-99.

### DIFF
--- a/ACPs/77-reinventing-subnets/README.md
+++ b/ACPs/77-reinventing-subnets/README.md
@@ -59,44 +59,11 @@ To validate an `AddressedCall` payload in Avalanche Warp Messaging, the `(chainI
 #### ConvertSubnetTx
 
 ```golang
-type ConvertSubnetTx struct {
-    // Metadata, inputs and outputs
-    BaseTx
-    // ID of the Subnet to transform
-    // Restrictions:
-    // - Must not be the Primary Network ID
-    Subnet ids.ID `json:"subnetID"`
-    // Chain where the Subnet manager lives
-    ChainID ids.ID `json:"chainID"`
-    // Address of the Subnet manager
-    Address []byte `json:"address"`
-    // Authorizes this conversion
-    SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
-}
-```
-
-Once this transaction is accepted, `AddSubnetValidatorTx` and `RemoveSubnetValidatorTx` are disabled on the Subnet. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward.
-
-### Recovering a Subnet
-
-In the event that a Subnet has no validators, a valid BLS multi-signature cannot be produced. This is the case immediately after a Subnet is created, and can arise later if all the Subnet's validators are removed from the validator set.
-
-A `RecoverSubnetTx` can be used in this situation to instantiate the Subnet's validator set using the Subnet's current `Owner`. In all other situations, the `Owner` key is powerless after a `ConvertSubnetTx` is issued. The only exception is for `Owner` rotation via the `TransferSubnetOwnershipTx`.
-
-Each Subnet Validator will be instantiated with a balance specified in the `Balance` field in each Subnet Validator. For a `RecoverSubnetTx` to be valid, each `Balance` must be >= the greater of 5 $AVAX or two weeks of the current fee. This prevents Subnet Validators from being added with too low of an initial balance where they become immediately delinquent based on the continuous fee mechanism defined below. A Subnet Validator can leave at any time before the initial $AVAX is consumed and claim the remaining balance to the `ChangeOwner` defined in the transaction.
-
-Note: An `EndTime` is specified for each Subnet Validator in the `RecoverSubnetTx` to enable a smooth transition to the Subnet Manager contract. The contract is not expected to manage the Validators defined in the `RecoverSubnetTx`.
-
-#### RecoverSubnetTx
-
-```golang
 type SubnetValidator struct {
     // Must be Ed25519 NodeID
     NodeID ids.NodeID `json:"nodeID"`
     // Weight of this validator used when sampling
     Weight uint64 `json:"weight"`
-    // When this validator will stop validating the Subnet
-    EndTime uint64 `json:"endTime"`
     // Initial balance for this validator
     Balance uint64 `json:"balance"`
     // [Signer] is the BLS key for this validator.
@@ -109,21 +76,74 @@ type SubnetValidator struct {
     ChangeOwner fx.Owner `json:"changeOwner"`
 }
 
-type RecoverSubnetTx struct {
+type ConvertSubnetTx struct {
     // Metadata, inputs and outputs
     BaseTx
-    // Validator set for the Subnet
+    // ID of the Subnet to transform
+    // Restrictions:
+    // - Must not be the Primary Network ID
+    Subnet ids.ID `json:"subnetID"`
+    // Chain where the Subnet manager lives
+    ChainID ids.ID `json:"chainID"`
+    // Address of the Subnet manager
+    Address []byte `json:"address"`
+    // Initial pay-as-you-go validators for the Subnet
     Validators []SubnetValidator `json:"validators"`
-    // Auth that will be allowing these validators into the network
+    // Authorizes this conversion
     SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
 }
+```
+
+Once this transaction is accepted, the `Owner` key is powerless. Additionally, `AddSubnetValidatorTx` and `RemoveSubnetValidatorTx` are disabled on the Subnet. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward.
+
+The `validationID` for validators added in `ConvertSubnetTx` is defined as the SHA256 hash of `(convertSubnetTxID,validatorIndex)`, where `validatorIndex` refers to the index into the `Validators` array of within the `ConvertSubnetTx`.
+
+The following serialization is defined as an `InitialSubnetValidator`:
+```text
++--------------+----------+-----------+
+|       nodeID : [32]byte |  32 bytes |
++--------------+----------+-----------+
+|       weight :   uint64 |   8 bytes |
++--------------+----------+-----------+
+| blsPublicKey : [48]byte |  48 bytes |
++--------------+----------+-----------+
+                          |  88 bytes |
+                          +-----------+
+```
+The following serialization is defined as the `SubnetConversionData`
+```text
++-------------------+--------------------------+--------------------------------------------------------------+
+| convertSubnetTxID :                 [32]byte |                                                     32 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+|    managerChainID :                 [32]byte |                                                     32 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+|    managerAddress :                   []byte |                                4 + len(managerAddress) bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+| initialValidators : []InitialSubnetValidator |                        4 + len(initialValidators) * 88 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+                                               | 72 + len(managerAddress) + len(initialValidators) * 88 bytes |
+                                               +--------------------------------------------------------------+
+```
+The `subnetConversionID` is defined as the SHA256 hash of the `SubnetConversionData` from a given `ConvertSubnetTx`.
+
+Once a `ConvertSubnetTx` is accepted, P-Chain validators will be willing to sign a `SubnetConversionMessage`, specified as an `AddressedCall` with a empty `originSenderAddress` with the following payload.
+```text
++--------------------+----------+----------+
+|            codecID :   uint16 |  2 bytes |
++--------------------+----------+----------+
+|             typeID :   uint32 |  4 bytes |
++--------------------+----------+----------+
+| subnetConversionID : [32]byte | 32 bytes |
++--------------------+----------+----------+
+                                | 38 bytes |
+                                +----------+
 ```
 
 ### Adding Subnet Validators
 
 #### RegisterSubnetValidatorTx
 
-A `RegisterSubnetValidatorTx` can be used to add a Subnet Validator.
+After a `ConvertSubnetTx` has been accepted, the `RegisterSubnetValidatorTx` can be used to add Subnet validators.
 
 ```golang
 type RegisterSubnetValidatorTx struct {
@@ -176,7 +196,7 @@ The `Message` field must be an `AddressedCall` with the payload:
 - `subnetID`, `nodeID`, `weight`, and `blsPublicKey` are for the Subnet Validator being added
 - `expiry` is the time after which this message is invalid. After the P-Chain timestamp is past `expiry`, this Avalanche Warp Message can no longer be used to add the `nodeID` to the validator set of `subnetID`.
 
-    `validationID` is defined as the SHA256 hash of the `Payload` of the `AddressedCall`. This SHA256 hash will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` is required to be no longer than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` with an `expiry` in the past can be flushed from the P-Chain's state.
+    `validationID` of validators added via `RegisterSubnetValidatorTx` is defined as the SHA256 hash of the `Payload` of the `AddressedCall`. This SHA256 hash will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` is required to be no longer than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` with an `expiry` in the past can be flushed from the P-Chain's state.
 
 Subnets are responsible for defining the procedure on how to retrieve the above information from prospective validators.
 
@@ -230,7 +250,7 @@ The `Message` field must be an `AddressedCall` with the payload:
 |      codecID :   uint16 |  2 bytes |
 +--------------+----------+----------+
 |       typeID :   uint32 |  4 bytes |
-+-----------+----------+----------+
++--------------+----------+----------+
 | validationID : [32]byte | 32 bytes |
 +--------------+----------+----------+
 |        nonce :   uint64 |  8 bytes |
@@ -256,6 +276,8 @@ The `Message` field must be an `AddressedCall` with the payload:
 Issuing a `SetSubnetValidatorWeightTx` with `Weight` of `0` will remove the Subnet Validator from the Subnet's validator set.
 
 All state related to the Subnet Validator being removed will be removed from the P-Chain's active state (BLS key, NodeID etc). This validator can be newly added to the Subnet's validator set using the `RegisterSubnetValidatorTx` flow.
+
+Since altering a Subnet's validator set after a `ConvertSubnetTx` requires a valid Warp message from the Subnet's current validator set, it is explicitly disallowed for a Subnet to remove it's only validator, which would render the Subnet unable to produce any further valid Warp messages. If a Subnet only has a single validator, it is invalid to set that validator's weight to 0 via a `SetSubnetValidatorWeightTx`.
 
 ### Disabling Subnet Validators
 
@@ -443,7 +465,6 @@ Any state execution changes must be coordinated through a mandatory upgrade. Imp
 
 - P-Chain
   - `ConvertSubnetTx`
-  - `RecoverSubnetTx`
   - `RegisterSubnetValidatorTx`
   - `SetSubnetValidatorWeightTx`
   - `DisableValidatorTx`


ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,74 @@ type SubnetValidator struct {
...
+
+Once a `ConvertSubnetTx` is accepted, P-Chain validators will be willing to sign a `SubnetConversionMessage`, specified as an `AddressedCall` with a empty `originSenderAddress` with the following payload.
+```text
++--------------------+----------+----------+
+|            codecID :   uint16 |  2 bytes |
++--------------------+----------+----------+
+|             typeID :   uint32 |  4 bytes |
++--------------------+----------+----------+
+| subnetConversionID : [32]byte | 32 bytes |
++--------------------+----------+----------+
```
minghinmatthewlam, line 137: should this `SubnetConversionMessage` include the `SubnetConversionData`? Or is that passed in separately to the validator manager contracts, in order to perform verification of subnet conversion ID

michaelkaplan13: Yeah exactly, where needed the pre-image of the `subnetConversionID` (which is the `SubnetConversionData`) can be provided and checked against the hash in the message.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,74 @@ type SubnetValidator struct {
...
+type ConvertSubnetTx struct {
     // Metadata, inputs and outputs
     BaseTx
-    // Validator set for the Subnet
+    // ID of the Subnet to transform
+    // Restrictions:
+    // - Must not be the Primary Network ID
+    Subnet ids.ID `json:"subnetID"`
+    // Chain where the Subnet manager lives
+    ChainID ids.ID `json:"chainID"`
```
meaghanfitzgerald, line 87:
```suggestion
    // BlockchainID where the L1 manager lives
    ChainID ids.ID `json:"chainID"`
```
ChainID (an integer set in genesis that has no functional use afaiaa) is often confused with _this instance_ of ChainID (the BlockchainID of the L1 derived from the txn hash of `CreateChainTx`).

michaelkaplan13: 👍 FWIW, the EVM `chainID` set in subnet-evm genesis files is used by those chains to enforce replay protection across any other EVM chains. This is why transactions sent from accounts on Ethereum can't be broadcasted on other EVM chains.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
+    Address []byte `json:"address"`
+    // Initial pay-as-you-go validators for the Subnet
     Validators []SubnetValidator `json:"validators"`
-    // Auth that will be allowing these validators into the network
+    // Authorizes this conversion
     SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
 }
 ```
 
+Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet. The only action that the `Owner` key is able to take is removing previously existing Subnet validators via `RemoveSubnetValidatorTx`. These previously existing validators otherwise will be removed automatically at their specified `EndTime`. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward, and the `Owner` key is powerless to modify anything related to these newly added validators.
```
meaghanfitzgerald, line 97:
```suggestion
Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet. The only action that the `Owner` key is able to take is removing previously existing Subnet validators via `RemoveSubnetValidatorTx`. Any previously existing validators added with the pre-ACP-77 workflow will otherwise be removed automatically at their specified [`EndTime`](https://github.com/ava-labs/avalanchego/blob/a1721541754f8ee23502b456af86fea8c766352a/vms/platformvm/txs/validator.go#L27). `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward, and the `Owner` key is powerless to modify anything related to these newly added validators.
```
wanted to make it clear that only pre-etna validators have a designated EndTime

ACPs/77-reinventing-subnets/README.md
```diff
@@ -443,13 +468,12 @@ Any state execution changes must be coordinated through a mandatory upgrade. Imp
...
 - P-Chain
   - `ConvertSubnetTx`
-  - `RecoverSubnetTx`
   - `RegisterSubnetValidatorTx`
   - `SetSubnetValidatorWeightTx`
   - `DisableValidatorTx`
   - `IncreaseBalanceTx`
 
-Once a Subnet issues a `ConvertSubnetTx`, `AddSubnetValidatorTx` and `RemoveSubnetValidatorTx` can no longer be used to modify that Subnet's validator set. A Subnet Validator added with an `AddSubnetValidatorTx` will continue to validate the Subnet until their `EndTime` is reached. After expiry, those validators (if they want to continue validating) must use the `RegisterSubnetValidatorTx` flow outlined in this ACP to register as a Subnet Validator.
+Once a Subnet issues a `ConvertSubnetTx`, `AddSubnetValidatorTx` can no longer be used to add validators to that Subnet's validator set. Any Subnet Validators previously added with an `AddSubnetValidatorTx` will continue to validate the Subnet until their `EndTime` is reached, or until removed by the Subnet `Owner` key via `RemoveSubnetValidatorTx`. After expiry or removal, those validators (if they want to continue validating) must use the `RegisterSubnetValidatorTx` flow outlined in this ACP to register as a Subnet Validator.
```
meaghanfitzgerald, line 476:
```suggestion
Once a Subnet issues a `ConvertSubnetTx`, `AddSubnetValidatorTx` can no longer be used to add validators to that Subnet's validator set. Any Subnet Validators previously added with an `AddSubnetValidatorTx` will continue to validate the Subnet until their [`EndTime`]([`EndTime`](https://github.com/ava-labs/avalanchego/blob/a1721541754f8ee23502b456af86fea8c766352a/vms/platformvm/txs/validator.go#L27)) is reached, or until removed by the Subnet `Owner` key via `RemoveSubnetValidatorTx`. After expiry or removal, those validators (if they want to continue validating) must use the `RegisterSubnetValidatorTx` flow outlined in this ACP to register as a Subnet Validator.
```

meaghanfitzgerald (approved):

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
-type RecoverSubnetTx struct {
+type ConvertSubnetTx struct {
     // Metadata, inputs and outputs
     BaseTx
-    // Validator set for the Subnet
+    // ID of the Subnet to transform
+    // Restrictions:
+    // - Must not be the Primary Network ID
+    Subnet ids.ID `json:"subnetID"`
+    // BlockchainID where the L1 manager lives
```
dhrubabasu, line 86: We've stuck to Subnet terminology everywhere else in the spec

```suggestion
    // BlockchainID where the Subnet manager lives
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
     SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
 }
 ```
 
+Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet. The only action that the `Owner` key is able to take is removing previously existing Subnet validators via `RemoveSubnetValidatorTx`. Any previously existing validators added with the pre-ACP-77 workflow will otherwise be removed automatically at their specified [`End`](https://github.com/ava-labs/avalanchego/blob/a1721541754f8ee23502b456af86fea8c766352a/vms/platformvm/txs/validator.go#L27) time. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward, and the `Owner` key is powerless to modify anything related to these newly added validators.
+
+The `validationID` for validators added in `ConvertSubnetTx` is defined as the SHA256 hash of `(convertSubnetTxID,validatorIndex)`, where `validatorIndex` refers to the index into the `Validators` array of within the `ConvertSubnetTx`.
+
+The following serialization is defined as an `InitialSubnetValidator`:
+```text
```
dhrubabasu, line 102: I've generally left a gap before and after a code block through the spec (I believe markdownlint complains about it)

ACPs/77-reinventing-subnets/README.md
```diff
@@ -257,6 +280,8 @@ Issuing a `SetSubnetValidatorWeightTx` with `Weight` of `0` will remove the Subn
 
 All state related to the Subnet Validator being removed will be removed from the P-Chain's active state (BLS key, NodeID etc). This validator can be newly added to the Subnet's validator set using the `RegisterSubnetValidatorTx` flow.
 
+Since altering a Subnet's validator set after a `ConvertSubnetTx` requires a valid Warp message from the Subnet's current validator set, it is explicitly disallowed for a Subnet to remove it's only validator, which would render the Subnet unable to produce any further valid Warp messages. If a Subnet only has a single validator, it is invalid to set that validator's weight to 0 via a `SetSubnetValidatorWeightTx`.
+
```
dhrubabasu, line 284: If the Subnet only has a single validator, that validator is explicitly signing that they want to be removed. I wonder if it would be a better design to garbage collect Subnets that have 0 validators than somewhat arbitrarily require that the last Validator to be removed is the only revival mechanism. If the last Validator stops paying the continuous fee, the Subnet is effectively dead. 

Relatedly, if 50% of a Subnet's weight stops paying the continuous fee, the Subnet is effectively dead. I believe the guidance there is to fork the Subnet with those Validators removed. Would that be a suitable answer for Subnets that get in this position?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
+|    managerAddress :                   []byte |                                4 + len(managerAddress) bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+| initialValidators : []InitialSubnetValidator |                        4 + len(initialValidators) * 88 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+                                               | 72 + len(managerAddress) + len(initialValidators) * 88 bytes |
+                                               +--------------------------------------------------------------+
+```
+The `subnetConversionID` is defined as the SHA256 hash of the `SubnetConversionData` from a given `ConvertSubnetTx`.
+
+Once a `ConvertSubnetTx` is accepted, P-Chain validators will be willing to sign a `SubnetConversionMessage`, specified as an `AddressedCall` with a empty `originSenderAddress` with the following payload.
```
dhrubabasu, line 129: It's unclear from the spec how `SubnetConversionData` is factored in here. How will Subnets consume `SubnetConversionData`?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
+| convertSubnetTxID :                 [32]byte |                                                     32 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+|    managerChainID :                 [32]byte |                                                     32 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+|    managerAddress :                   []byte |                                4 + len(managerAddress) bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+| initialValidators : []InitialSubnetValidator |                        4 + len(initialValidators) * 88 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+                                               | 72 + len(managerAddress) + len(initialValidators) * 88 bytes |
+                                               +--------------------------------------------------------------+
```
dhrubabasu, line 125: `initialValidators` feels a little misleading since the Subnet can have other validators from `AddSubnetValidatorTx`. I think dropping initial would be best since these are strictly the validators added in the conversion tx.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
+    Address []byte `json:"address"`
+    // Initial pay-as-you-go validators for the Subnet
     Validators []SubnetValidator `json:"validators"`
-    // Auth that will be allowing these validators into the network
+    // Authorizes this conversion
     SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
 }
 ```
 
+Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet. The only action that the `Owner` key is able to take is removing previously existing Subnet validators via `RemoveSubnetValidatorTx`. Any previously existing validators added with the pre-ACP-77 workflow will otherwise be removed automatically at their specified [`End`](https://github.com/ava-labs/avalanchego/blob/a1721541754f8ee23502b456af86fea8c766352a/vms/platformvm/txs/validator.go#L27) time. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward, and the `Owner` key is powerless to modify anything related to these newly added validators.
```
dhrubabasu, line 97:
```suggestion
Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet going forward. The only action that the `Owner` key is able to take is removing any Subnet validators that were added using `AddSubnetValidatorTx` previously via `RemoveSubnetValidatorTx`. Unless removed by the `Owner` key, note that any Subnet Validators added previously with an `AddSubnetValidatorTx` will continue to validate the Subnet until their `EndTime` is reached. Once all previously added Subnet Validators are no longer in the validator set, the `Owner` key is powerless. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward.
```

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Validator manager contracts will need to learn of the validators specified in the `ConvertSubnetTx` in order to be able to remove them in the future. The `SubnetConversionData` includes all of those validators. It's expected for validator manager contracts to be provided the `SubnetConversionData` that is the pre-image of the `subnetConversionID` provided via a `SubnetConversionMessage`. 

We could have the `SubnetConversionData` itself be the `SubnetConversionMessage`, but it can be arbitrarily large due to the `validators` array, and the P-Chain will need to be willing to sign the `SubnetConversionMessage` in perpetuity for a given Subnet, so we should keep it to a fix size (similar to tracking Subnet IDs and Chain IDs on the P-Chain today).

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Would defer to @StephenButtolph here.

But IIUC, validators stopping paying the continuous fee or directly disabling themselves on the P-Chain is functionally different from the final validators removing themselves because they could always re-up/re-enable themselves to revive the Subnet in that case, where if they were fully removed then there is zero recovery path.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Added

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,77 @@ type SubnetValidator struct {
...
+|    managerChainID :                 [32]byte |                                                     32 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+|    managerAddress :                   []byte |                                4 + len(managerAddress) bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+| initialValidators : []InitialSubnetValidator |                        4 + len(initialValidators) * 88 bytes |
++-------------------+--------------------------+--------------------------------------------------------------+
+                                               | 72 + len(managerAddress) + len(initialValidators) * 88 bytes |
+                                               +--------------------------------------------------------------+
+```
+The `subnetConversionID` is defined as the SHA256 hash of the `SubnetConversionData` from a given `ConvertSubnetTx`.
```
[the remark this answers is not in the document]
michaelkaplan13: Good call, moved it down

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Updated 👍

ACPs/77-reinventing-subnets/README.md
StephenButtolph: - I think it might make sense to move the "If the last validator leaves then the subnet is deleted" to a followup... It would fit very easily into a future change if we were to just disallow it in this ACP.
- I don't think that the subnet should be removed if too much of the stake becomes deactivated (As the subnet isn't known to be dead, because balances can be increased without a Warp message).

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Got it- I was missing that nuance of the pre-image

ACPs/77-reinventing-subnets/README.md
dhrubabasu: I agree that Subnets shouldn't be deleted if too much of the stake becomes deactivated. For explicit removals via `SetSubnetValidatorWeightTx` with `Weight=0` is a different story. 

Disallowing the last one to be removed seems like an odd design pattern to me since the validator is explicitly signing off on its own removal. Shouldn't it be allowed that they are able to remove themselves? 

The `RecoverSubnetTx` was designed to guard against this since we could kick-back recovery path to the `Owner`. Instead of re-introducing that tx (which had other issues as @michaelkaplan13 noted), I think it's reasonable to say the Subnet can no longer be revived. If anybody wants to revive the Subnet, you'll need to fork the state and create a new Subnet.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -443,13 +472,12 @@ Any state execution changes must be coordinated through a mandatory upgrade. Imp
...
   - `DisableValidatorTx`
   - `IncreaseBalanceTx`
 
-Once a Subnet issues a `ConvertSubnetTx`, `AddSubnetValidatorTx` and `RemoveSubnetValidatorTx` can no longer be used to modify that Subnet's validator set. A Subnet Validator added with an `AddSubnetValidatorTx` will continue to validate the Subnet until their `EndTime` is reached. After expiry, those validators (if they want to continue validating) must use the `RegisterSubnetValidatorTx` flow outlined in this ACP to register as a Subnet Validator.
+Once a Subnet issues a `ConvertSubnetTx`, `AddSubnetValidatorTx` can no longer be used to add validators to that Subnet's validator set. Any Subnet Validators previously added with an `AddSubnetValidatorTx` will continue to validate the Subnet until their [`End`](https://github.com/ava-labs/avalanchego/blob/a1721541754f8ee23502b456af86fea8c766352a/vms/platformvm/txs/validator.go#L27) time is reached, or until removed by the Subnet `Owner` key via `RemoveSubnetValidatorTx`. After expiry or removal, those validators (if they want to continue validating) must use the `RegisterSubnetValidatorTx` flow outlined in this ACP to register as a Subnet Validator.
```
dhrubabasu, line 480: I know I left this in here but I think we can remove this. It doesn't really belong in the Backwards Compatibility section since you have to use a new feature to disable old features.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -172,11 +199,11 @@ The `Message` field must be an `AddressedCall` with the payload:
 ```
 
 - `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
-- `typeID` is the payload type identifier and is `0x00000000` for this transaction
+- `typeID` is the payload type identifier and is `0x00000001` for this transaction
 - `subnetID`, `nodeID`, `weight`, and `blsPublicKey` are for the Subnet Validator being added
-- `expiry` is the time after which this message is invalid. After the P-Chain timestamp is past `expiry`, this Avalanche Warp Message can no longer be used to add the `nodeID` to the validator set of `subnetID`.
+- `expiry` is the time at which this message becomes invalid. As of a P-Chain timestamp `expiry` or greater, this Avalanche Warp Message can no longer be used to add the `nodeID` to the validator set of `subnetID`.
```
dhrubabasu, line 204:
```suggestion
- `expiry` is the time at which this message becomes invalid. As of a P-Chain timestamp `>= expiry`, this Avalanche Warp Message can no longer be used to add the `nodeID` to the validator set of `subnetID`.
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,81 @@ type SubnetValidator struct {
...
+|             typeID :   uint32 |  4 bytes |
++--------------------+----------+----------+
+| subnetConversionID : [32]byte | 32 bytes |
++--------------------+----------+----------+
+                                | 38 bytes |
+                                +----------+
+```
+
+- `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
+- `typeID` is the payload type identifier and is `0x00000000` for this transaction
```
dhrubabasu, line 146:
```suggestion
- `typeID` is the payload type identifier and is `0x00000000` for this message
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,81 @@ type SubnetValidator struct {
...
+- `subnetConversionID` is the SHA256 hash of the `SubnetConversionData` from a given `ConvertSubnetTx`.
```
dhrubabasu, line 147:
```suggestion
- `subnetConversionID` is the SHA256 hash of the `SubnetConversionData` from a given `ConvertSubnetTx`
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,81 @@ type SubnetValidator struct {
...
++-------------------+--------------------------+-------------------------------------------------------+
+|    managerAddress :                   []byte |                         4 + len(managerAddress) bytes |
++-------------------+--------------------------+-------------------------------------------------------+
+|        validators :          []ValidatorData |                        4 + len(validators) * 88 bytes |
++-------------------+--------------------------+-------------------------------------------------------+
+                                               | 72 + len(managerAddress) + len(validators) * 88 bytes |
+                                               +-------------------------------------------------------+
+```
+
+Once a `ConvertSubnetTx` is accepted, P-Chain validators will be willing to sign a `SubnetConversionMessage`, specified as an `AddressedCall` with a empty `originSenderAddress` with the following payload.
```
dhrubabasu, line 131:
```suggestion
Once a `ConvertSubnetTx` is accepted, P-Chain validators must be willing to sign an `AddressedCall` with `sourceChainID` set to the P-Chain ID and the `sourceAddress` set to an empty byte array. The payload of the `AddressedCall` must be:
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,81 @@ type SubnetValidator struct {
...
     Validators []SubnetValidator `json:"validators"`
-    // Auth that will be allowing these validators into the network
+    // Authorizes this conversion
     SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
 }
 ```
 
+Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet going forward. The only action that the `Owner` key is able to take is removing any Subnet validators that were added using `AddSubnetValidatorTx` previously via `RemoveSubnetValidatorTx`. Unless removed by the `Owner` key, note that any Subnet Validators added previously with an `AddSubnetValidatorTx` will continue to validate the Subnet until their `EndTime` is reached. Once all Subnet Validators added with `AddSubnetValidatorTx` are no longer in the validator set, the `Owner` key is powerless. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward.
+
+The `validationID` for validators added in `ConvertSubnetTx` is defined as the SHA256 hash of `(convertSubnetTxID,validatorIndex)`, where `validatorIndex` refers to the index into the `Validators` array of within the `ConvertSubnetTx`.
```
dhrubabasu, line 99: I think we should be a bit more exact with the SHA256 hash. I believe it will be using [`ids.Prefix`](https://github.com/ava-labs/avalanchego/blob/master/ids/id.go#L97).

```suggestion
The `validationID` for validators added through `ConvertSubnetTx` is defined as the SHA256 hash of the 40 byte concatenation of the 32 byte `convertSubnetTxID` and the 8 byte `validatorIndex` (index in the `Validators` array within the transaction).
```

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: I think this is mostly a UX question for this extreme edge case.

In the event that the only Subnet validator wants to kill the Subnet forever, it seems like it would be functionally equivalent for them to just stop/delete their node for that Subnet as compared to setting their weight to 0 on the P-Chain.

On the other hand though, explicitly disallowing the destructive action on the P-chain could act as a defensive-in-depth guard rail for a single validator accidentally signing to remove themselves when they don't actually want to kill the Subnet forever.

I would prefer to play it on the safer side of not allowing the more destructive action for now (which would be more likely to occur accidentally immediately following activation as developers are figuring out how to properly integrate), and if the number of effectively "dead" Subnets in the future ever becomes an issue, then it could be changes with a separate ACP.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Good call, removed now.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,81 @@ type SubnetValidator struct {
...
++-------------------+--------------------------+-------------------------------------------------------+
+|    managerAddress :                   []byte |                         4 + len(managerAddress) bytes |
++-------------------+--------------------------+-------------------------------------------------------+
+|        validators :          []ValidatorData |                        4 + len(validators) * 88 bytes |
++-------------------+--------------------------+-------------------------------------------------------+
+                                               | 72 + len(managerAddress) + len(validators) * 88 bytes |
+                                               +-------------------------------------------------------+
+```
+
+Once a `ConvertSubnetTx` is accepted, P-Chain validators must be willing to sign an `AddressedCall` with `sourceChainID` set to the P-Chain ID and the `sourceAddress` set to an empty byte array. The payload of the `AddressedCall` must be:
```
iansuvak, line 131: With the rewording of this line we've lost the only reference to `SubnetConversionMessage` in this PR although it's still in the title of the PR and presumably will be referenced in the `avalanchego` code implementation. Should we add it back?

dhrubabasu: I prefer to not name them in the ACP if it's not referenced again in the ACP (the other AddressedCall payloads don't have designated names)

ACPs/77-reinventing-subnets/README.md
dhrubabasu: Having a guard rail for this makes sense to me. However, I'm not sure the P-Chain itself is the right place for it. Forcing a validator to stay in the validator set when they're explicitly signing that they want to be removed still seems like a strange design with the only benefit being this guard rail. While signing for removal is functionally equivalent to simply stopping/deleting their node, it doesn't remove them from the validator set meaning their NodeID is still associated with the Subnet. This could potentially cause regulatory concerns as well (iirc some Primary Network Validators were concerned about not being able to early eject before the EndTime)

Should this be handled on the Subnet Manager contract side instead? If you want to remove the last validator of the set, you have to remove it through a different method than the standard method. I guess this could be viewed kind of like a "Are you sure you want to delete this Subnet?" prompt. That should avoid the accidental signature concern you raised

ACPs/77-reinventing-subnets/README.md
```diff
@@ -109,21 +76,81 @@ type SubnetValidator struct {
...
     SubnetAuth verify.Verifiable `json:"subnetAuthorization"`
 }
 ```
 
+Once this transaction is accepted, `AddSubnetValidatorTx` is disabled on the Subnet going forward. The only action that the `Owner` key is able to take is removing any Subnet validators that were added using `AddSubnetValidatorTx` previously via `RemoveSubnetValidatorTx`. Unless removed by the `Owner` key, any Subnet Validators added previously with an `AddSubnetValidatorTx` will continue to validate the Subnet until their [`End`](https://github.com/ava-labs/avalanchego/blob/a1721541754f8ee23502b456af86fea8c766352a/vms/platformvm/txs/validator.go#L27) time is reached. Once all Subnet Validators added with `AddSubnetValidatorTx` are no longer in the validator set, the `Owner` key is powerless. `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` must be used to manage the Subnet's validator set going forward.
+
+The `validationID` for validators added through `ConvertSubnetTx` is defined as the SHA256 hash of the 40 bytes resulting from concatenating the 32 byte `convertSubnetTxID` with the 8 byte `validatorIndex` (index in the `Validators` array within the transaction).
```
iansuvak, line 99: The `validatorIndex` here being 8 byte value implies that it's a uint64 but the spec for `SubnetConversionData` below specifies the length of validatorData being of length 4 implying uint32. 

Is the intention that this is a uint32 packed as a uint64 to make the full length of the SHA256 pre-image 40?

michaelkaplan13: Great catch. All lengths should be 4 bytes (uint32), so the `validatorIndex` should be a uint32 as well, and the pre-image should be 36 bytes. 

Updated now

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: To play devil's advocate here, what's the gained benefit from not explicitly enforcing it on the P-chain when it's functionally equivalent? I don't think there are any regulatory concerns regarding a completely halted chain that will never produce any more blocks or activity, and node IDs are freely rotated. Validator manager contract wise, the final validator actually would never be able to completely remove themselves because the process requires posting an ack of the removal on the P-chain back on the L1, which would at that point not be able to produce a new block.

To me it seems like removals of the final validator are far more likely to be an accident than intentional, and even when intentional the benefit/motivation to do so is not very clear.

dhrubabasu: The main benefit is a validator isn't locked into the Subnet's validator set on the P-chain. A validator is able to willingly turn off or delete their node. With this guardrail, the P-chain would prohibit them from removing themselves.

> Validator manager contract wise, the final validator actually would never be able to completely remove themselves because the process requires posting an ack of the removal on the P-chain back on the L1, which would at that point not be able to produce a new block.

The canonical validator set of a Subnet lives on the P-chain. If a validator is removed on the P-chain, they are removed from the Subnet's canonical validator set and will no longer participate in consensus. Sending the ack back to the L1 isn't required to stop participating.

> To me it seems like removals of the final validator are far more likely to be an accident than intentional, and even when intentional the benefit/motivation to do so is not very clear.

I think my earlier suggestion of requiring removal through a different method on the contract would sufficiently prevent accidental removals without locking in any validators on the P-chain.

> I don't think there are any regulatory concerns regarding a completely halted chain that will never produce any more blocks or activity, and node IDs are freely rotated.

This isn't true, otherwise this wouldn't be a guardrail. The validator can come back online via the `IncreaseBalanceTx` which would revive the Subnet. It's not guaranteed that the validator will never participate in the Subnet again, potentially causing regulatory concerns. 

Fwiw I don't think the regulatory stuff is that big of a point. I was just reminded of some concerns Primary Network Validators brought up when thinking about this guardrail. I'm mainly disagreeing with placing a somewhat arbitrary restriction in the P-chain that isn't required for safety/liveness of the P-chain. 

This feels to me like a Subnet-level guardrail, not a Primary Network-level guardrail. For example, a Subnet may want two PAYG validators always attached to the Subnet instead of just one. If they want to do that, that would need to live in the Subnet contract.

ACPs/77-reinventing-subnets/README.md
dhrubabasu: If we're using [ids.Prefix](https://github.com/ava-labs/avalanchego/blob/master/ids/id.go#L97), the uint32 will get cast to a uint64. See UTXOID: https://github.com/ava-labs/avalanchego/blob/master/vms/components/avax/utxo_id.go#L45-L51

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Yeah, fair enough. I was just talking with @StephenButtolph about this topic, his take on it is slightly different:
- If the only Subnet validator is removed on the P-Chain (permanently killing the Subnet), then all of the other Subnet state should also be cleaned up and removed from the P-Chain.
- How to properly remove the entire state of a Subnet is outside the scope of the ACP-77 and the Etna upgrade.
- Allowing the only Subnet validator to be removed should be deferred to a later ACP, which includes the full Subnet state clean up. Until then, permanently killing Subnets without removing them from state will be accumulating tech debt.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Hm, in that case it would be 40 bytes with the 8 byte validator index first then?

iansuvak: Yeah the ordering of the fields in the spec would need to change if using `ids.Prefix`

ACPs/77-reinventing-subnets/README.md
dhrubabasu: As far as I understand, cleaning up and removing dead Subnets from the P-chain state would not constitute an execution level change. This would mean it is not required to scope out how to properly remove a dead Subnet from the P-chain state in ACP-77, the initial Etna release, or a future ACP. ACP-77 could go forward without having any restrictions on the removal of the last PAYG validator. AvalancheGo can resolve the issue of cleaning up a dead Subnet at any later point in time. This tech debt feels like a minor problem since `CreateSubnetTx` can currently be issued without any cleanup of a Subnet's state, even when the Subnet is functionally dead (e.g. https://subnets.avax.network/subnets/2TSYtJmLdmBJCTcEP6ZeQ3d7urg1GbrKtVMMdd7nmz46z22PSC).

If we include this restriction in Etna, the minimum point at which we can enable Subnet validators to remove themselves if they are the last validator is in the F upgrade. On the other hand, we can not have this restriction in Etna and clean up the state in a minor release of AvalancheGo.

Given the above, I think we should not introduce this restriction in Etna. Would love to hear @StephenButtolph's thoughts on this.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Thoughts on keeping the (convertSubnetTxID,index) ordering, and not adding the 4 zero bytes to the pre-image?

Changing the serialization for the sake of using an existing helper feels a bit odd. Could always add a new helper if needed

ACPs/77-reinventing-subnets/README.md
StephenButtolph: You're right, cleaning up the subnet state wouldn't need to be a consensus decision.

- I feel like allowing deletion of the last SoV _could_ brick a subnet.
- I feel like there is no clear reason to allow the last SoV to leave the subnet.

Specifically:
> The main benefit is a validator isn't locked into the Subnet's validator set on the P-chain. A validator is able to willingly turn off or delete their node. With this guardrail, the P-chain would prohibit them from removing themselves.

All the validator needs to do to stop validating a subnet, is remove the subnetID from their `--track-subnets` flag. In fact, it's possible to add any validator to any subnet. Additionally, doing so could be done with a BLS key that the actual owner of the nodeID doesn't have access to. So if we actually care about this case, we need to restructure the BLS PoP.

All-in-all, I feel like just disallowing this behavior is the most future-proof thing to do. We can circle back to how we want to handle this edge case in the future.

dhrubabasu: > In fact, it's possible to add any validator to any subnet. Doing so can be done with a BLS key that the actual owner of the nodeID doesn't have access to. So, if we actually care about this case, we'd need to restructure the BLS PoP.

The BLS PoP point is a good one. A subtle difference here is that we'd be restricting removal even when the NodeID is signing off on it by virtue of being the last validator in the Subnet.

I've thought about it more and re-read the discussion here. I definitely agree that allowing deletion of the last SoV could brick a Subnet. Forking and re-creating a new Subnet at this point is sub-optimal to say the least (especially without any tooling). There are strong merits (as both @StephenButtolph and @michaelkaplan13 pointed out) on having the guardrail live in the P-chain until at least F. Personally, I think the end-vision of this design would be removing this restriction (in a follow-up ACP), cleaning up the Subnet state entirely, and moving the guardrail to the Subnet Manager contract level.

Would it make sense to include some language to add color here? What do you guys think about:

```suggestion
If all Subnet Validators are removed, there are no valid Warp messages that can be produced to register new Subnet Validators through `RegisterSubnetValidatorTx`. With no validators, block production will halt and the Subnet is unrecoverable. To serve as a guardrail against this situation, a Subnet Validator removal will be considered invalid if the validator being removed is the last one. A future ACP can remove this guardrail as users get more familiar with the new Subnet mechanics and tooling matures to fork a Subnet.
```

Nuttymoon: Just a side note here, one way for a Subnet to protect against death by empty validator set is to set its manager address on another chain, e.g. the C-Chain. This is the design we have chosen for Suzaku, where all contracts will be hosted on the C-Chain at it is the more decentralized and secure of the ecosystem today.

dhrubabasu: The same problem applies. With an empty validator set, block production will halt. The only way to make progress is have a non-empty validator set which isn't guarded against by not using the Subnet for the manager contract. The only way to guard against it is to either prohibit an empty validator set on the P-Chain or on the manager contract.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Unless anyone is strongly opposed, I suggest we keep this serialization specification as it is currently since it avoid the extra zero bytes and IMO has a more intuitive ordering.

dhrubabasu: I don't really have an opinion here tbh. `ids.Prefix` was floated in earlier convos so I figured we'd just go with that.

If we end up using `ids.Prefix` we should update the spec to match that. If we don't, what you have now is perfectly valid.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -257,6 +284,7 @@ Issuing a `SetSubnetValidatorWeightTx` with `Weight` of `0` will remove the Subn
 
 All state related to the Subnet Validator being removed will be removed from the P-Chain's active state (BLS key, NodeID etc). This validator can be newly added to the Subnet's validator set using the `RegisterSubnetValidatorTx` flow.
 
+If all Subnet Validators are removed, there are no valid Warp messages that can be produced to register new Subnet Validators through `RegisterSubnetValidatorTx`. With no validators, block production will halt and the Subnet is unrecoverable. To serve as a guardrail against this situation, a Subnet Validator removal will be considered invalid if the validator being removed is the last one. A future ACP can remove this guardrail as users get more familiar with the new Subnet mechanics and tooling matures to fork a Subnet.
```
dhrubabasu, line 287: I messed up the spacing in my suggestion, needs an empty space after

```suggestion
If all Subnet Validators are removed, there are no valid Warp messages that can be produced to register new Subnet Validators through `RegisterSubnetValidatorTx`. With no validators, block production will halt and the Subnet is unrecoverable. To serve as a guardrail against this situation, a Subnet Validator removal will be considered invalid if the validator being removed is the last one. A future ACP can remove this guardrail as users get more familiar with the new Subnet mechanics and tooling matures to fork a Subnet.

```

dhrubabasu (approved):
Nice cleanup, this design is a lot better to me than the `RecoverSubnetTx` flow with `EndTime`s

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: I think @Nuttymoon's point is that even if it is guarded against in the contracts (which I agree it should be), if the validator manager is on the L1 with a single validator itself, that validator could sign the arbitrary message to remove themselves even if the contract doesn't allow for emitting that message on-chain.

But if it's on the C-Chain and the contract doesn't allow emitting that message, then the single validator can't remove themselves in any way.

dhrubabasu: Good point, I glossed over that when I first ready @Nuttymoon's comment. Putting it on the C-Chain does guard against the Subnet's validator set side-stepping the rules of the manager contract.

PR #148, opened 2024-09-30
[ACP-77] Adjust nomenclature due to renaming of Avalanche Subnets -> Avalanche L1s

meaghanfitzgerald:
As stated in this Avalanche Academy [Guide](https://academy.avax.network/guide/etna-changes#subnets-are-now-called-l1s): _Avalanche Subnets have been rebranded as Avalanche L1s._ 

To avoid confusion, the spec has been revised to account for this, including the following changes: 

```
ConvertSubnetTx -> ConvertL1Tx
RegisterSubnetValidatorTx -> RegisterL1ValidatorTx
RegisterSubnetValidatorMessage -> RegisterL1ValidatorMessage
SetSubnetValidatorWeightTx -> SetL1ValidatorWeightTx
DisableValidatorTx -> DisableL1ValidatorTx
SubnetValidatorRegistrationMessage -> L1ValidatorRegistrationMessage
IncreaseBalanceTx -> IncreaseL1ValidatorBalanceTx 
```

### To Review:

[ReadMe Preview](https://github.com/meaghanfitzgerald/ACPs/blob/nomenclature-change/ACPs/77-reinventing-subnets/README.md)

### Nuance:

- `Subnet validator` continues to refer to all validators created and controlled by Subnet Owners pre-Etna Upgrade, controllable by the `AddSubnetValidatorTx` and `RemoveSubnetValidatorTx`. 
- `CreateSubnetTx` is a pre-Etna transaction whose functionality and nomenclature will not change within this spec. 
- `SubnetID` is the transaction hash produced by a successful `CreateSubnetTx` and a parameter required by some of the new transaction types. This will not be renamed.

PR #149, opened 2024-10-01
[ACP-77] Message and transaction specification updates and regrouping

michaelkaplan13:
This PR makes a number of relatively small tweaks to the ACP-77 messages and transaction specifications to help improve simplicity of implementation and UX. Thanks to @StephenButtolph for calling out many of the possible improvements as he's worked to provide an implementation of this ACP.

- Updates the `ConvertSubnetTx` specification to only include a BLS proof-of-possession signature, which doesn't require its own type ID.
- Renames `ChangeOwner` to `RemainingBalanceOwner`
- Adds the `RemainingBalanceOwner` to the `RegisterSubnetValidatorMessage` to prevent possible griefing attacks where the deliverer of the message on the P-Chain can set their own value for this field. This field is removed from the `RegisterSubnetValidatorTx`.
- Adds a `codecID` to the start of the `SubnetConversionData` to allow for standard serialization.
- Replaces the `convertSubnetTxID` in `SubnetConversionData` with the `SubnetID`, and similarly redefines the `validationID` for validators added in a `ConvertSubnetTx`. The `SubnetID` is already a unique hash, so using the transaction ID was unnecessary.
- Changes the `signer` in the `RegisterSubnetValidatorTx` to be just a BLS signature, since the BLS public key is already included in the Warp message within the transaction.
- Moves `weight` to be the last field of the Warp messages, for consistency with other field orderings.
- Removes the requirement of 5 AVAX for the initial balance of a Subnet validator, since it could be worked around by disabling the validator and then increasing its balance to a lesser amount. There are no initial balance requirements, a Subnet Validator could start with a balance of 0 and be immediately `inactive`.
- Adds the `disableOwner` to the `RegisterSubnetValidatorMessage`. This owner is able to disable the validator directly on the P-Chain, rather than requiring an Ed25519 signature from the node ID. This helps improve UX and removes the requirement that 32-byte node IDs are used. Support for 32-byte node IDs can be added in a future upgrade, and the message formats specified here will not need modification since they use variable length node IDs.
- Removes the duplicate `SubnetValidatorWeightUpdateMessage` serializations with different type IDs.
- Assigns the `typeID`s for the new Warp messages payloads to be sequential in the order that they are used for new Subnets.

The ACP document is also reorganized to group all of the Warp message payload serializations together in one section, and all of the new P-Chain transactions together in another. This is purely to help make the document easier to navigate as it has grown larger.


ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,194 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
 
 ## Specification
 
-### Separate Subnet Validators from Primary Network Validators
+At a high-level, Subnets will now be able to manage their validator sets external to the P-Chain by setting the blockchain ID and address of their "validator manager". This validator manager is able to send Avalanche Warp Messages to the P-Chain in order to modify the validator set registered for the Subnet. In response, the P-Chain will send Avalanche Warp Message back to validator managers to notify them of validator set changes that have taken effect. Subnet Validators will no longer be required to validate the Primary Network, which removes the requirement to stake 2000 $AVAX to be a Subnet Validator. Further, Subnet Validators will only be required to sync the P-Chain (not X/C-Chain) in order to track validator set changes and support cross-Subnet communication. To maintain a Subnet Validator on the P-Chain, a continuous fee must be paid in $AVAX as specified later in this ACP.
```
minghinmatthewlam, line 35: nit: why is Validator capitalized in some places but not others, is it solely for "Subnet Validators"?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,194 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
 
-To provide maximal flexibility for Permissionless Subnets, the BLS multisignature approach in Avalanche Warp Messaging (AWM) is re-used to approve modifications to the Subnet's validator set. Using the `(chainID, address)` pair defined in the `ConvertSubnetTx`, a Warp Message with an [`AddressedCall`](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm/warp/payload#addressedcall) payload can be constructed.
+The following serialization is defined as a `ValidatorData`:
 
-To validate an `AddressedCall` payload in Avalanche Warp Messaging, the `(chainID, address)` pair is used to lookup the validators of `chainID` and verify that the BLS multi-sig includes a quorum (set to 67%) of the Subnet's validator set. The P-Chain will use Warp Messages with `AddressedCall` payloads to support modifications of the Subnet's validator set using this `(chainID, address)` pair (using the `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` defined later in the specification).
+| Field | Type | Size |
+| -: | -: | -: |
+| `nodeID` | `[]byte` | 4 + len(`nodeID`) bytes |
```
minghinmatthewlam, line 62: is it worth adding the explanation of why nodeID is variable length

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,194 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+#### `RegisterSubnetValidatorMessage`
+
+The `RegisterSubnetValidatorMessage` is sent from validator managers to the P-Chain to inform the P-Chain of additions to Subnet's validator set.
+
+The following is the serialization of a `PChainOwner`:
+
+| Field | Type | Size |
+| -: | -: | -: |
+| `threshold` | `uint32` | 4 bytes |
+| `addresses` | `[][20]byte` | 4 + len(`addresses`) * 20 bytes |
```
minghinmatthewlam, line 107: can you add explanation of these fields, similar to other messages. I'm wondering what threshold is, and why address is known to be 20 bytes, compared to others that are variable length.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,194 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+| `disableOwner` | `PChainOwner` | 8 + len(`addresses`) * 20 bytes |
+| `weight` | `uint64` | 8 bytes |
+| | | 122 + len(`nodeID`) + (len(`addresses1`) + len(`addresses2`)) * 20 bytes |
+
+- `codecID` is the codec version used to serialize the payload, and is hardcoded to `0x0000`
+- `typeID` is the payload type identifier and is `0x00000001` for this payload
+- `subnetID`, `nodeID`, `weight`, and `blsPublicKey` are for the Subnet Validator being added
+- `expiry` is the time at which this message becomes invalid. As of a P-Chain timestamp `>= expiry`, this Avalanche Warp Message can no longer be used to add the `nodeID` to the validator set of `subnetID`
+- `remainingBalanceOwner` is the P-Chain owner where leftover $AVAX from the Subnet Validator's Balance will be issued to when this validator it is removed from the validator set.
+- `disableOwner` is only the P-Chain owner allowed to disable the validator using `DisableSubnetValidatorTx`, specified below.
```
minghinmatthewlam, line 130:
```suggestion
- `disableOwner` is the only P-Chain owner allowed to disable the validator using `DisableSubnetValidatorTx`, specified below.
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +281,65 @@ An EVM Subnet may choose to implement this step like so:
...
 
 ```golang
-type DisableValidatorTx struct {
+type DisableSubnetValidatorTx struct {
     // Metadata, inputs and outputs
     BaseTx
     // ID corresponding to the validator
     ValidationID ids.ID `json:"validationID"`
-    // Ed25519 Signature on [TxID]
-    Signature []byte `json:"signature"`
```
StephenButtolph, line 308: Now that this can be a multi-sig, we need to add a `DisableAuth verify.Verifiable` here.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Yeah exactly, it has only been for "Subnet Validators". I'm not 100% sure if it should even be capitalized then, but I think it's still consistent to that end

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Good call. I added a note above the message specification since it applies to all of the node IDs the messages contain.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Added. 

All P-Chain addresses are represented as `ids.ShortID`, which are 20 bytes. I left that part out because I don't think it's super relevant here, similar to how we don't explain why BLS signatures are known to be 96 bytes, etc.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Thank for calling that out. Added now

ACPs/77-reinventing-subnets/README.md
meaghanfitzgerald: I made all mentions of validator lower case in #148

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
-
-`SetSubnetValidatorWeightTx` is used to modify the voting weight of a Subnet Validator. For this transaction to be valid, 67% of the current Subnet Validator weight must sign the Warp message included in `Message`. Applications of this transaction could include:
-
-- Increase the voting weight of a Subnet Validator if a delegation is made on the Subnet
-- Increase the voting weight of a Subnet Validator if the stake amount is increased (by staking rewards for example)
-- Decrease the voting weight of a misbehaving Subnet Validator
-- Remove an inactive Subnet Validator
-
-Since there are no `EndTime`s enforced by the P-Chain, Subnets must use this transaction to set an expired validator's weight to `0`. When a validator's weight is set to `0`, they will be removed from the validator set. After the transaction is accepted, any $AVAX remaining in the `Balance` for the Subnet Validator being removed will be returned to the `ChangeOwner` defined in the `RegisterSubnetValidatorTx`.
+`SetSubnetValidatorWeightTx` is used to modify the voting weight of a Subnet Validator. The specifcation of this transaction is:
```
cam-schultz, line 300:
```suggestion
`SetSubnetValidatorWeightTx` is used to modify the voting weight of a Subnet Validator. The specfication of this transaction is:
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
-| validationID : [32]byte | 32 bytes |
-+--------------+----------+----------+
-|        nonce :   uint64 |  8 bytes |
-+--------------+----------+----------+
-|       weight :   uint64 |  8 bytes |
-+--------------+----------+----------+
-                          | 54 bytes |
-                          +----------+
-```
+There is no requirement of any interaction on the Subnet to issue this transaction. For this transaction to be valid, a signature from the `DisableOwner` specified for this validator in its `RegisterSubnetValidatorTx`. Any remaining $AVAX in the Subnet Validator's `Balance` will be issued to the `RemaningBalanceOwner` defined when this validator was added to the validator set.
```
cam-schultz, line 339:
```suggestion
There is no requirement of any interaction on the Subnet to issue this transaction. For this transaction to be valid, a signature from the `DisableOwner` specified for this validator in its `RegisterSubnetValidatorTx`. Any remaining $AVAX in the Subnet Validator's `Balance` will be issued to the `RemainingBalanceOwner` defined when this validator was added to the validator set.
```

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Think we both had typos here...correcting now

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+    //       - len(Addresses) == 0 if `Threshold` is 0.
+    //       - len(Addresses) >= `Threshold`
+    //       - The values in Addresses to be sorted in ascending order.
+    Addresses []ids.ShortID `json:"addresses"`
+}
+
 type SubnetValidator struct {
-    // Must be Ed25519 NodeID
+    // NodeID of this validator
     NodeID ids.NodeID `json:"nodeID"`
```
StephenButtolph, line 216: Should this also be a variable length nodeID?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -94,116 +240,38 @@ type ConvertSubnetTx struct {
...
 
-    `validationID` of validators added via `RegisterSubnetValidatorTx` is defined as the SHA256 hash of the `Payload` of the `AddressedCall`. This SHA256 hash will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` is required to be no more than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` corresponding to an expired timestamp can be flushed from the P-Chain's state.
+This `validationID` will be used for replay protection. Used `validationID`s will be stored on the P-Chain. If a `RegisterSubnetValidatorTx`'s `validationID` has already been used, the transaction will be considered invalid. To prevent storing an unbounded number of `validationID`s, the `expiry` of the `RegisterSubnetValidatorMessage` is required to be no more than 48 hours in the future of the time the transaction is issued on the P-Chain. Any `validationIDs` corresponding to an expired timestamp can be flushed from the P-Chain's state.
```
StephenButtolph, line 282: 48h is likely too much. Currently avalanchego has it set to 24h, it might need to be even smaller.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: I guess that would make future upgrades to support new node ID types less disruptive? Makes sense to me.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Totally defer to whatever you think is safe from a P-Chain perspective. 

UX wise, I think as low as 30 minutes is sufficient.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+| `validationID` | `[32]byte` | 32 bytes |
+| `registered` | `bool` | 1 byte  | 
+| | | 39 bytes |
+
+- `codecID` is the codec version used to serialize the payload, and is hardcoded to `0x0000`
+- `typeID` is the payload type identifier and is `0x00000002` for this message
+- `validationID` identifies the validator for the message
+- `registered` is a boolean representing the status of the `validationID`. If true, the `validationID` corresponds to a validator in the current validator set. If false, the `validationID` does not correspond to a validator in the current validator set, and never will in the future.
+
+#### `SubnetValidatorWeightMessage`
```
Nuttymoon, line 159: Why change this from `SubnetValidatorWeightUpdateMessage`? I do believe that the latter makes more sense in regards to `SubnetValidatorRegistrationMessage`.

michaelkaplan13: The thinking was that this message format is also used for messages from the P-Chain back to the L1, and the P-Chain is only ever willing to sign for the latest validator weight and nonce, as opposed to signing for each individual weight update.

Nuttymoon: Makes sense!

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
 
-### Managing Subnet Validator Balance
+Note that this does not modify a Subnet's total staking weight. This transaction marks the Subnet Validator as inactive, but does not remove it from the Subnet's validator set. Inactive Subnet Validators can re-activate at any time by increasing their balance with an `IncreaseSubnetValidatorWeightTx`.
 
-#### IncreaseBalanceTx
+Subnet creators should be aware that there is no notion of `MinStakeDuration` that is enforced by the P-Chain. It is expected that Subnets who choose to enforce a `MinStakeDuration` will lock the validator's Stake for the Subnet's desired `MinStakeDuration`.
 
-This transaction can be issued by anybody to add additional $AVAX to the `Balance` for `NodeID` validating `Subnet`. If the Subnet Validator corresponding to `ValidationID` is currently inactive (`Balance` was exhausted or `DisableValidatorTx` was issued), this transaction will move them back to the active validator set.
+#### `IncreaseSubnetValidatorWeightTx`
```
dhrubabasu, line 347: Why was this transaction renamed from `IncreaseBalanceTx` to `IncreaseSubnetValidatorWeightTx`? Does it increase the weight of the validator now?

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+### P-Chain Warp Message Payloads
 
-#### Bootstrapping Subnet Nodes
+To support modifying a Subnet's validator set on the P-Chain via Warp messages, Warp message verification support will be added to the [`PlatformVM`](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm). Similar to Warp messages received by the C-Chain, in order for a Warp message to be considered valid by the P-Chain, at least 67% of the `sourceChainID`'s validator set must have partcipated in the aggregate BLS signature.
```
dhrubabasu, line 39: I believe it's 67% of the _weight_ rather than the number of validators. I re-worded it a bit to clarify that point and include that this threshold could be modified on a per-Subnet basis in the future.

```suggestion
To enable management of a Subnet's validator set externally to the P-Chain, Warp message verification will be added to the [`PlatformVM`](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm). For a Warp message to be considered valid by the P-Chain, at least 67% of the `sourceChainID`'s weight must have participated in the aggregate BLS signature. This is equivalent to the threshold set for the C-Chain. A future ACP may be proposed to support modification of this threshold on a per-Subnet basis.
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+- `RegisterSubnetValidatorTx`
+- `SetSubnetValidatorWeightTx`
+- `DisableSubnetValidatorTx`
+- `IncreaseSubnetValidatorWeightTx`
 
-To validate an `AddressedCall` payload in Avalanche Warp Messaging, the `(chainID, address)` pair is used to lookup the validators of `chainID` and verify that the BLS multi-sig includes a quorum (set to 67%) of the Subnet's validator set. The P-Chain will use Warp Messages with `AddressedCall` payloads to support modifications of the Subnet's validator set using this `(chainID, address)` pair (using the `RegisterSubnetValidatorTx` and `SetSubnetValidatorWeightTx` defined later in the specification).
+#### `ConvertSubnetTx`
 
-#### ConvertSubnetTx
+After a `CreateSubnetTx` is issued, a `ConvertSubnetTx` may be issued by the `Owner` to explicitly convert a Subnet from permissioned to permissionless. This transaction will set the `(chainID, address)` pair that will manage the Subnet's validator set going forward.
```
dhrubabasu, line 197: Re-worded a bit to put the state transition prior to the validation criteria

```suggestion
To convert a Subnet from permissioned to permissionless, a `ConvertSubnetTx` must be issued to set the `(chainID, address)` pair that will manage the Subnet's validator set going forward. The `Owner` key defined in `CreateSubnetTx` must provide a signature to authorize this conversion.
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
 
 ```golang
 type SetSubnetValidatorWeightTx struct {
     // Metadata, inputs and outputs
     BaseTx
-    // AddressedCall with Payload:
-    //   - ValidationID (SHA256 of the AddressedCall Payload of the RegisterSubnetValidatorTx adding the validator)
-    //   - Nonce
-    //   - Weight
+    // A SubnetValidatorWeightMessage payload
```
dhrubabasu, line 306: This pattern is a _lot_ clearer than it was in the original spec. Thanks for taking the time to cleanup ❤️

ACPs/77-reinventing-subnets/README.md
```diff
@@ -383,9 +359,23 @@ type IncreaseBalanceTx struct {
...
+After this ACP is activated, the P-Chain will no longer support staking of any assets other than $AVAX for the Primary Network. The P-Chain will no longer support distribution of staking rewards for Subnets. All staking-related operations for Subnet Validation must be managed by the Subnet's validator manager. The P-Chain simply requires a continuous fee per Subnet Validator. If a Subnet would like to manage their Validator's balances on the P-Chain, it can cover the cost for all Subnet Validators by posting the $AVAX balance on the P-Chain. Subnets can implement any mechanism they want to pay the continuous fee charged by the P-Chain for its participants.
```
dhrubabasu, line 378: great catch

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
 
-#### IncreaseBalanceTx
+Subnet creators should be aware that there is no notion of `MinStakeDuration` that is enforced by the P-Chain. It is expected that Subnets who choose to enforce a `MinStakeDuration` will lock the validator's Stake for the Subnet's desired `MinStakeDuration`.
 
-This transaction can be issued by anybody to add additional $AVAX to the `Balance` for `NodeID` validating `Subnet`. If the Subnet Validator corresponding to `ValidationID` is currently inactive (`Balance` was exhausted or `DisableValidatorTx` was issued), this transaction will move them back to the active validator set.
+#### `IncreaseSubnetValidatorWeightTx`
 
-Note: The $AVAX added to `Balance` can be claimed at any time by the Subnet Validator using `DisableValidatorTx`.
+Subnet Validators are required to maintain a non-zero balance used to pay the continuous fee on the P-Chain in order to be considered active. The `IncreaseSubnetValidatorWeightTx` can be used by anybody to add additional $AVAX to the `Balance` to a Subnet Validator. The specification for this transaction is:
```
dhrubabasu, line 349: nit of all nits: can we standardize either `The specification *for* this transaction is` or `The specification *of* this transaction is`? we use each twice in this PR currently

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
-| validationID : [32]byte | 32 bytes |
-+--------------+----------+----------+
-|        nonce :   uint64 |  8 bytes |
-+--------------+----------+----------+
-|       weight :   uint64 |  8 bytes |
-+--------------+----------+----------+
-                          | 54 bytes |
-                          +----------+
-```
+There is no requirement of any interaction on the Subnet to issue this transaction. For this transaction to be valid, a signature from the `DisableOwner` specified for this validator in its `RegisterSubnetValidatorTx`. Any remaining $AVAX in the Subnet Validator's `Balance` will be issued to the `RemainingBalanceOwner` defined when this validator was added to the validator set.
```
dhrubabasu, line 339: I think we can skip saying "there isn't any interaction on the Subnet" here since we mention it below

```suggestion
The `DisableOwner` specified for this validator must sign the transaction. Any unspent $AVAX in the Subnet Validator's `Balance` will be issued in a single UTXO to the `RemainingBalanceOwner` for this validator. Recall that both `DisableOwner` and `RemainingBalanceOwner` are specified when the validator is first added to the Subnet's validator set (in either `ConvertSubnetTx` or `RegisterSubnetValidatorTx`).
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
 
-- `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
-- `typeID` is the payload type identifier and is `0x00000004` for this message
-- `validationID` is the SHA256 of the `Payload` of the `AddressedCall` in the `RegisterSubnetValidatorTx` adding the validator to the Subnet's validator set
-- `nonce` is the latest nonce stored with the `validationID` on the P-Chain
-- `weight` is the current weight of the Subnet Validator with `validationID`
+The expected path for full removal from a Subnet's validator set is via a `SetSubnetValidatorWeightTx` with weight `0` which requires a Warp message produced by the Subnet. However, the ability to reclaim the balance allotted to a validator without authorization from the Subnet is critical for failed Subnets.
```
dhrubabasu, line 341: "The expected path" language suggests that this transaction is a full removal, cleaned that up a bit

```suggestion
For full removal from a Subnet's validator set, a `SetSubnetValidatorWeightTx`  must be issued with weight `0`. To do so, a Warp message is required from the Subnet's manager. However, to support the ability to claim the unspent `Balance` for a Subnet Validator without authorization is critical for failed Subnets.
```

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: Yikes, no this was a typo. Thanks for catching it, should be `IncreaseSubnetValidatorBalanceTx`, to be consistent with the transaction naming of the other new transactions for managing subnet validators.

ACPs/77-reinventing-subnets/README.md
michaelkaplan13: 🫡

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
 
-#### Bootstrapping Subnet Nodes
+To enable management of a Subnet's validator set externally to the P-Chain, Warp message verification will be added to the [`PlatformVM`](https://github.com/ava-labs/avalanchego/tree/master/vms/platformvm). For a Warp message to be considered valid by the P-Chain, at least 67% of the `sourceChainID`'s weight must have participated in the aggregate BLS signature. This is equivalent to the threshold set for the C-Chain. A future ACP may be proposed to support modification of this threshold on a per-Subnet basis.
 
-Bootstrapping a node/validator is the process of securely recreating the latest state of the blockchain locally. At the end of this process, the local state of a node/validator must be in sync with the local state of other virtuous nodes/validators. The node/validator can then verify new incoming transactions and reach consensus with other nodes/validators.
+For messages sent from the P-Chain for a given Subnet, only that Subnet's validators must be willing to provide signatures, rather than the entire Primary Network validator set. This optimization is possible because all Subnet Validators will still sync the P-Chain.
```
dhrubabasu, line 41:
```suggestion
To verify messages produced by the P-Chain, typically the entire Primary Network's validator set needs to participate in the aggregate BLS signature. However, this can cause a great deal of network bandwidth, increasing latency. An optimization can be made to only require that a quorum of the _Subnet's_ validator set needs to participate in the aggregate BLS signature for messages that are produced by the P-Chain and consumed by the Subnet.
```

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
 
 ## Specification
 
-### Separate Subnet Validators from Primary Network Validators
+At a high-level, Subnets will now be able to manage their validator sets external to the P-Chain by setting the blockchain ID and address of their "validator manager". This validator manager is able to send Avalanche Warp Messages to the P-Chain in order to modify the validator set registered for the Subnet. In response, the P-Chain will send Avalanche Warp Message back to validator managers to notify them of validator set changes that have taken effect. Subnet Validators will no longer be required to validate the Primary Network, which removes the requirement to stake 2000 $AVAX to be a Subnet Validator. Further, Subnet Validators will only be required to sync the P-Chain (not X/C-Chain) in order to track validator set changes and support cross-Subnet communication. To maintain a Subnet Validator on the P-Chain, a continuous fee must be paid in $AVAX as specified later in this ACP.
```
dhrubabasu, line 35:
```suggestion
At a high-level, Subnets can manage their validator sets externally to the P-Chain by setting the blockchain ID and address of their _validator manager_. The P-Chain will consume Warp messages that modify the Subnet's validator set. To confirm modification of the Subnet's validator set, the P-Chain will also produce Warp messages. Subnet Validators will no longer be required to validate the Primary Network, removing the 2000 $AVAX stake requirement to be a Subnet Validator. In place of this requirement, a continuous fee denominated in $AVAX is introduced to maintain an active Subnet Validator. Subnet Validators are only required to sync the P-Chain (not X/C-Chain) in order to track validator set changes and support cross-Subnet communication.
```

dhrubabasu (commented):
Last set of comments!

I think we should stray away from "send" verbage since the P-Chain or the validator manager won't be automatically sending the Warp messages to each other. I replaced it with "produce/consume" verbage but feel free to modify my suggestions.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -213,167 +289,67 @@ An EVM Subnet may choose to implement this step like so:
...
 
-- `codecID` is the codec version used to serialize the payload and is hardcoded to `0x0000`
-- `typeID` is the payload type identifier and is `0x00000004` for this message
-- `validationID` is the SHA256 of the `Payload` of the `AddressedCall` in the `RegisterSubnetValidatorTx` adding the validator to the Subnet's validator set
-- `nonce` is the latest nonce stored with the `validationID` on the P-Chain
-- `weight` is the current weight of the Subnet Validator with `validationID`
+For full removal from a Subnet's validator set, a `SetSubnetValidatorWeightTx`  must be issued with weight `0`. To do so, a Warp message is required from the Subnet's manager. However, to support the ability to claim the unspent `Balance` for a Subnet Validator without authorization is critical for failed Subnets.
```
dhrubabasu, line 341:
```suggestion
For full removal from a Subnet's validator set, a `SetSubnetValidatorWeightTx` must be issued with weight `0`. To do so, a Warp message is required from the Subnet's manager. However, to support the ability to claim the unspent `Balance` for a Subnet Validator without authorization is critical for failed Subnets.
```

michaelkaplan13:
> Last set of comments!
> 
> I think we should stray away from "send" verbage since the P-Chain or the validator manager won't be automatically sending the Warp messages to each other. I replaced it with "produce/consume" verbage in my suggestions.
> 
> Feel free to modify my suggestions as you see fit!

Yeah good call. I made the updates now. Only tweaks were to avoid including motivation/reasoning in the specification section, and making it clear that consumers of messages from the P-Chain are intended to be validator managers.

ACPs/77-reinventing-subnets/README.md
```diff
@@ -32,48 +32,202 @@ With no Elastic Subnets live on Mainnet, it is clear that Permissionless Subnets
...
+The following is the serialization of a `PChainOwner`:
+
+| Field | Type | Size |
+| -: | -: | -: |
+| `threshold` | `uint32` | 4 bytes |
+| `addresses` | `[][20]byte` | 4 + len(`addresses`) * 20 bytes |
+| | | 8 + len(`addresses`) * 20 bytes |
+
+- `threshold` is the number of `addresses` that must provide a signature for the `PChainOwner` to authorize an action.
+- In order to be a valid:
```
dhrubabasu, line 113:
```suggestion
- Validation criteria:
```

dhrubabasu (approved):
This PR makes ACP-77 much clearer, great work 🏆

michaelkaplan13:
> This PR makes ACP-77 much clearer, great work 🏆

@StephenButtolph gets the credit for suggested changes and improvements here 🙏

PR #151, opened 2024-10-10
[ACP 150] Use current block's P-Chain height for ICM signature verification

iansuvak:
```
ACP: 150
Title: Use current block's P-Chain height for ICM signature verification
Author(s): Ian Suvak (https://github.com/ian-suvak)
Discussions-To: 
Status: Proposed
Track: Standards
```

## Abstract

Proposes that VMs use the P-Chain height of the current block being built instead of the parent block when verifying aggregated signatures of Avalanche Interchain Messages (ICM). This will allow for a more reliable way to determine which validators should participate in signing the message, and remove unnecessary waiting periods.

[Rendered](https://github.com/iansuvak/ACPs/tree/acp-150-use-current-pchain-height/ACPs/150-icm-use-current-block-pchain-height)


ACPs/151-icm-use-current-block-pchain-height/README.md
```diff
@@ -0,0 +1,38 @@
+| ACP           | 151                                                             |
+| :------------ | :-------------------------------------------------------------- |
+| **Title**     | Use current block P-Chain height for ICM signature verification |
+| **Author(s)** | Ian Suvak ([@iansuvak](https://github.com/iansuvak))            |
+| **Status**    | Proposed                                                        |
+| **Track**     | Standards                                                       |
+
+## Abstract
+
+Proposes that VMs use the P-Chain height of the current block being built instead of the parent block when verifying aggregated signatures of Avalanche Interchain Messages (ICM). This will allow for a more reliable way to determine which validators should participate in signing the message, and remove unnecessary waiting periods.
```
michaelkaplan13, line 10:
```suggestion
Proposes that the ProposerVM uses the P-Chain height of the current block being built instead of the parent block when verifying aggregated signatures of Avalanche Interchain Messages (ICM). This will allow for a more reliable way to determine which validators should participate in signing the message, and remove unnecessary waiting periods.
```

ACPs/151-icm-use-current-block-pchain-height/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Motivation
+
+Currently ICM messages use P-Chain height of the parent block when verifying messages for inclusion in the current block. Using the parent's P-Chain height is necessary for safe block building purposes but not for ICM message verification.
```
michaelkaplan13, line 14:
```suggestion
Currently the ProposerVM uses the P-Chain height of the parent block when verifying ICM messages in the current block. Using the parent block's P-Chain height is necessary for verifying the proposer and reaching consensus on the current block, but it is not necessary not for verifying ICM messages within the block.
```

ACPs/151-icm-use-current-block-pchain-height/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Using the P-Chain height of the current block being built would make operations using ICM messages to modify the validator set, such as ones specified in [ACP-77](../77-reinventing-subnets/README.md) be verifiable sooner and more reliably. Currently at least two new P-Chain blocks need to be produced after the relevant state change for it to be reflected for purposes of ICM aggregate signature verification.
+
+## Specification
+
+[Block Context](https://github.com/ava-labs/avalanchego/blob/d2e9d12ed2a1b6581b8fd414cbfb89a6cfa64551/snow/engine/snowman/block/block_context_vm.go#L14) containing `PChainHeight` field is being passed to the VMs building the inner block, and is later being used by the VMs to fetch the canonical validator set for verification of ICM aggregated signatures. 
+
+The height currently passed in is the P-Chain height of the parent block which is the same height used by the ProposerVM to verify the proposer. This is necessary for proposer verification but not for verifying the state of this block so should be populated with the P-Chain height of the block being currently built instead.
```
michaelkaplan13, line 22:
```suggestion
The [block context](https://github.com/ava-labs/avalanchego/blob/d2e9d12ed2a1b6581b8fd414cbfb89a6cfa64551/snow/engine/snowman/block/block_context_vm.go#L14) contains a `PChainHeight` field that is passed from the ProposerVM to the inner VMs building the block. It is later used by the inner VMs to fetch the canonical validator set for verification of ICM aggregated signatures.

The `PChainHeight` currently passed in by the ProposerVM is the P-Chain height of the parent block. The proposed change is to instead have the ProposerVM pass in the P-Chain height of the current block.
```

The motivation already covers why this is safe and why it's useful, so think it's more concise to just include the exact specification of the change here.

ACPs/151-icm-use-current-block-pchain-height/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This change requires an upgrade to make sure that all validators verifying the validity of the ICM messages use the same P-Chain height and therefore the same validator set. Prior to activation nodes should continue to use P-Chain height of the parent block.
+
+## Reference Implementation
+
+A full reference implementation has not been provided yet. It must be provided prior to this ACP being considered "Implementable". 
+
+## Security Considerations
+
+ProposerVM needs to use the parent block's P-Chain height to verify proposers for security reasons but we don't have such restrictions for verifying ICM message validity in the current block being built, therefore this should be a safe change.
```
michaelkaplan13, line 34:
```suggestion
ProposerVM needs to use the parent block's P-Chain height to verify proposers for security reasons but we don't have such restrictions for verifying ICM message validity in the current block being built. Therefore, this should be a safe change.
```

michaelkaplan13 (commented):
Makes sense to me, just have a few clarification suggestions 🙏

ACPs/151-icm-use-current-block-pchain-height/README.md
iansuvak: ProposerVM in this case just passes in the context but doesn't do the verification itself, right? The real proposal here is that ProposerVM *passes* the current block's p-chain height as the context for verification of state to the inner VM so that the inner VM can use that for ICM verification. 

ProposerVM doesn't control what the inner VM does with the context once it's passed in right. I'm not sure what is the best way to word this though.

ACPs/151-icm-use-current-block-pchain-height/README.md
iansuvak: I like the second part of the suggestion but I have the same comment as above, re: wording implying that the ProposerVM does ICM message verification. My current wording is worse since it implies that ICM messages verify themselves 😓 . 

Should I just have the proposal / title of the ACP be `Use current block's P-Chain height for block state verification`.

And then mention ICM as the existing use-case in the motivation.

ACPs/151-icm-use-current-block-pchain-height/README.md
michaelkaplan13: Ah yeah, fair enough!

Technically this is only a change to the ProposerVM, which I was trying to get across here, but that of course impacts any VMs that use the ProposerVM

michaelkaplan13: How about that updated suggestion?

ACPs/151-icm-use-current-block-pchain-height/README.md
michaelkaplan13: > Should I just have the proposal / title of the ACP be Use current block's P-Chain height for block state verification. And then mention ICM as the existing use-case in the motivation.

I hadn't thought of that, but yeah it makes sense to me. Inner VMs could use this P-Chain for other purposes in the future hypothetically. 

I also updated the suggested wording here to reflect what you called out. Great point

michaelkaplan13 (approved):
🙌

PR #170, opened 2024-12-19
[ACP-99] Specify a standard `ACP99Manager` abstract contract

cam-schultz:
Makes the following changes as discussed [here](https://github.com/avalanche-foundation/ACPs/discussions/165#discussioncomment-11512285):
- Removes "Security Modules" from the specification
- Removes error definitions from the specification
- Specifies a unified `ACP99Manager` `abstract contract` (an `interface` is not used so that methods may be marked `internal`)

Additionally, this PR
- Removes getters for all active validations from the proposed interface
- Removes uptime parameters from the proposed interface
- Updates types to match those in use by Ava Labs' validator manager [contracts](https://github.com/ava-labs/icm-contracts/tree/main/contracts/validator-manager)
- Points to Suzaku's and Ava Labs' implementations as "example integrations" rather than reference implementations. A reference implementation in `ava-labs/icm-contracts` is in development.

### DIFF
--- a/ACPs/99-validatorsetmanager-contract/README.md
+++ b/ACPs/99-validatorsetmanager-contract/README.md
@@ -8,7 +8,7 @@
 
 ## Abstract
 
-Define (i) a reference interface for a minimal Validator Manager Solidity smart contract to be deployed on any Avalanche EVM chain, as well as (ii) a modular architecture to easily plug in custom “security modules” on top of this contract (e.g. to implement a PoS Subnet).
+Define a reference interface for a minimal Validator Manager Solidity smart contract to be deployed on any Avalanche EVM chain.
 
 This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets). It depends on it to be marked as `Implementable`.
 
@@ -18,16 +18,14 @@ This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https:/
 
 On each validator set change, the P-Chain is willing to sign an `AddressedCall` to notify any on-chain program tracking the validator set. On-chain programs must be able to interpret this message, so they can trigger the appropriate action. The 2 kinds of `AddressedCall`s [defined in ACP-77](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#p-chain-warp-message-payloads) are `L1ValidatorRegistrationMessage` and `L1ValidatorWeightMessage`.
 
-Given these assumptions and the fact that most of the active blockchains on Avalanche mainnet are EVM-based, we propose defining a reference implementation for a Solidity smart contract that can:
+Given these assumptions and the fact that most of the active blockchains on Avalanche mainnet are EVM-based, we propose `IACP99Manager` as the standard Solidity interface that can:
 
-1. Hold relevant information about the current Subnet validator set, as well as historical information
+1. Hold relevant information about the current Subnet validator set
 2. Send validator set updates to the P-Chain by generating `AdressedCall`s defined in ACP-77
 3. Correctly update the validator set by interpreting notification messages received from the P-Chain
-4. Be easily extended with custom “security modules” to implement any security model (e.g. PoS). Those modules have to implement the `IACP99SecurityModule` interface and will be called by the `ACP99Manager` contract upon validator set updates.
+4. Be easily integrated into applications implementing various security models (e.g. Proof-of-Stake).
 
-Having an audited and open-source reference implementation freely available will contribute to lowering the cost of launching Subnets on Avalanche.
-
-Once deployed, the `ACP99Manager` contract will be used as the `Address` in the [`ConvertSubnetToL1Tx`](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#convertsubnettol1tx).
+Once deployed, the `IACP99Manager` implementation contract will be used as the `Address` in the [`ConvertSubnetToL1Tx`](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#convertsubnettol1tx).
 
 ## Specification
 
@@ -38,14 +36,34 @@ Once deployed, the `ACP99Manager` contract will be used as the `Address` in the 
 Here is the proposed interface for the `IACP99Manager` contract:
 
 ```solidity
+/**
+ * @notice Description of the conversion data used to convert
+ * a subnet to an L1 on the P-Chain.
+ * This data is the pre-image of a hash that is authenticated by the P-Chain
+ * and verified by the Validator Manager.
+ */
+struct ConversionData {
+    bytes32 l1ID;
+    bytes32 validatorManagerBlockchainID;
+    address validatorManagerAddress;
+    InitialValidator[] initialValidators;
+}
+
+/// @notice Specifies an initial validator, used in the conversion data.
+struct InitialValidator {
+    bytes nodeID;
+    bytes blsPublicKey;
+    uint64 weight;
+}
+
 /// @notice L1 validation status
-enum ValidationStatus {
-    Registering,
+enum ValidatorStatus {
+    Unknown,
+    PendingAdded,
     Active,
-    Updating,
-    Removing,
+    PendingRemoved,
     Completed,
-    Expired
+    Invalidated
 }
 
 /**
@@ -58,47 +76,23 @@ struct PChainOwner {
 }
 
 /**
- * @notice L1 validation
- * @param status The validation status
- * @param nodeID The NodeID of the validator
- * @param startTime The start time of the validation
- * @param endTime The end time of the validation
- * @param periods The list of validation periods.
- * The index is the nonce associated with the weight update.
- * @param activeSeconds The time during which the validator was active during this validation
- * @param uptimeSeconds The uptime of the validator for this validation
+ * @notice Contains the active state of a Validation
+ * @notice status is the validation status
+ * @notice nodeID is The NodeID of the validator
+ * @notice startingWeight is the weight of the validator at the time of registration
+ * @notice messageNonce is the current weight update nonce
+ * @notice weight is the current weight of the validator
+ * @notice startTime The start time of the validation
+ * @notice endTime The end time of the validation
  */
 struct Validation {
     ValidationStatus status;
-    bytes32 nodeID;
-    ValidationPeriod[] periods;
-}
-
-/**
- * @notice L1 validation period
- * @param weight The weight of the validator during the period
- * @param startTime The start time of the validation period
- * @param endTime The end time of the validation period (only ≠ 0 when the period is over)
- */
-struct ValidationPeriod {
+    bytes nodeID;
+    uint64 startingWeight;
+    uint64 messageNonce;
     uint64 weight;
     uint64 startTime;
     uint64 endTime;
-    uint64 uptimeSeconds;
-}
-
-/**
- * @notice Information about a validator's uptime
- * @param activeSeconds The total number of seconds the validator was active
- * @param uptimeSeconds The total number of seconds the validator was online
- * @param activeWeightSeconds The total weight x seconds the validator was active
- * @param uptimeWeightSeconds The total weight x seconds the validator was online
- */
-struct ValidatorUptimeInfo {
-    uint64 activeSeconds;
-    uint64 uptimeSeconds;
-    uint256 activeWeightSeconds;
-    uint256 uptimeWeightSeconds;
 }
 
 /*
@@ -106,8 +100,6 @@ struct ValidatorUptimeInfo {
  * @notice The IACP99Manager interface is the interface for the ACP99Manager contract.
  */
 interface IACP99Manager {
-    /// @notice Emitted when the security module address is set
-    event SetSecurityModule(address indexed securityModule);
     /// @notice Emitted when an initial validator is registered
     event RegisterInitialValidator(
         bytes32 indexed nodeID, bytes32 indexed validationID, uint64 weight
@@ -136,68 +128,16 @@ interface IACP99Manager {
         bytes32 indexed nodeID, bytes32 indexed validationID, uint64 nonce, uint64 weight
     );
 
-    error ACP99Manager__ValidatorSetAlreadyInitialized();
-    error ACP99Manager__InvalidConversionID(bytes32 conversionID, bytes32 messageConversionID);
-    error ACP99Manager__InvalidManagerBlockchainID(
-        bytes32 managerBlockchainID, bytes32 conversionBlockchainID
-    );
-    error ACP99Manager__InvalidManagerAddress(address managerAddress, address conversionAddress);
-    error ACP99Manager__ZeroAddressSecurityModule();
-    error ACP99Manager__OnlySecurityModule(address sender, address securityModule);
-    error ACP99Manager__InvalidExpiry(uint64 expiry, uint256 timestamp);
-    error ACP99Manager__ZeroNodeID();
-    error ACP99Manager__NodeAlreadyValidator(bytes nodeID);
-    error ACP99Manager__InvalidPChainOwnerThreshold(uint256 threshold, uint256 addressesLength);
-    error ACP99Manager__PChainOwnerAddressesNotSorted();
-    error ACP99Manager__InvalidSignatureLength(uint256 length);
-    error ACP99Manager__InvalidValidationID(bytes32 validationID);
-    error ACP99Manager__InvalidWarpMessage();
-    error ACP99Manager__InvalidSourceChainID(bytes32 sourceChainID);
-    error ACP99Manager__InvalidOriginSenderAddress(address originSenderAddress);
-    error ACP99Manager__InvalidRegistration();
-    error ACP99Manager__NodeIDNotActiveValidator(bytes nodeID);
-    error ACP99Manager__InvalidUptimeValidationID(bytes32 validationID);
-    error ACP99Manager__InvalidSetL1ValidatorWeightNonce(uint64 nonce, uint64 currentNonce);
-
     /// @notice Get the ID of the Subnet tied to this manager
     function subnetID() external view returns (bytes32);
-
-    /// @notice Get the address of the security module attached to this manager
-    function getSecurityModule() external view returns (address);
 
     /// @notice Get the validation details for a given validation ID
     function getValidation(
         bytes32 validationID
     ) external view returns (Validation memory);
 
-    /// @notice Get the uptime information for a given validation ID
-    function getValidationUptimeInfo(
-        bytes32 validationID
-    ) external view returns (ValidatorUptimeInfo memory);
-
-    /// @notice Get an L1 validator's active validation ID
-    function getValidatorActiveValidation(
-        bytes memory nodeID
-    ) external view returns (bytes32);
-
-    /// @notice Get the current L1 validator set (list of NodeIDs)
-    function getActiveValidatorSet() external view returns (bytes32[] memory);
-
     /// @notice Get the total weight of the current L1 validator set
     function l1TotalWeight() external view returns (uint64);
-
-    /// @notice Get the list of message IDs associated with an L1 validator
-    function getValidatorValidations(
-        bytes memory nodeID
-    ) external view returns (bytes32[] memory);
-
-    /**
-     * @notice Set the address of the security module attached to this manager
-     * @param securityModule_ The address of the security module
-     */
-    function setSecurityModule(
-        address securityModule_
-    ) external;
 
     /**
      * @notice Verifies and sets the initial validator set for the chain through a P-Chain
@@ -208,10 +148,11 @@ interface IACP99Manager {
     function initializeValidatorSet(
         ConversionData calldata conversionData,
         uint32 messsageIndex
-    ) external;
+    ) internal;
 
     /**
-     * @notice Initiate a validator registration by issuing a RegisterL1ValidatorTx Warp message
+     * @notice Initiate a validator registration by issuing a RegisterL1ValidatorTx Warp message. The validator should
+     * not be considered active until completeValidatorRegistration is called.
      * @param nodeID The ID of the node to add to the L1
      * @param blsPublicKey The BLS public key of the validator
      * @param registrationExpiry The time after which this message is invalid
@@ -226,64 +167,41 @@ interface IACP99Manager {
         PChainOwner memory remainingBalanceOwner,
         PChainOwner memory disableOwner,
         uint64 weight
-    ) external returns (bytes32);
-
-    /**
-     * @notice Resubmits a validator registration message to be sent to P-Chain.
-     * Only necessary if the original message can't be delivered due to validator churn.
-     * @param validationID The validationID attached to the registration message
-     */
-    function resendValidatorRegistrationMessage(
-        bytes32 validationID
-    ) external returns (bytes32);
+    ) internal returns (bytes32);
 
     /**
      * @notice Completes the validator registration process by returning an acknowledgement of the registration of a
-     * validationID from the P-Chain.
+     * validationID from the P-Chain. The validator should not be considered active until this method is called.
      * @param messageIndex The index of the Warp message to be received providing the acknowledgement.
      */
     function completeValidatorRegistration(
         uint32 messageIndex
-    ) external;
-
-    /**
-     * @notice Updates the uptime of a validator for a given validation ID
-     * @param nodeID The ID of the node to update the uptime for
-     * @param messageIndex The index of the Warp message to be received providing the uptime proof
-     */
-    function updateUptime(
-        bytes memory nodeID,
-        uint32 messageIndex
-    ) external returns (ValidatorUptimeInfo memory);
+    ) internal returns (bytes32);
 
     /**
      * @notice Initiate a validator weight update by issuing a SetL1ValidatorWeightTx Warp message.
-     * If the weight is 0, this initiates the removal of the validator from the L1. An uptime proof can be
-     * included. This proof might be required to claim validator rewards (handled by the security module).
+     * If the weight is 0, this initiates the removal of the validator from the L1. The validator weight change
+     * should not have any effect until completeValidatorWeightUpdate is called.
      * @param nodeID The ID of the node to modify
      * @param weight The new weight of the node on the L1
-     * @param includesUptimeProof Whether the uptime proof is included in the message
-     * @param messageIndex The index of the Warp message containing the uptime proof
      */
     function initiateValidatorWeightUpdate(
-        bytes memory nodeID,
-        uint64 weight,
-        bool includesUptimeProof,
-        uint32 messageIndex
-    ) external;
+        bytes32 validationID,
+        uint64 weight
+    ) internal returns (uint64);
 
     /**
      * @notice Completes the validator weight update process by returning an acknowledgement of the weight update of a
-     * validationID from the P-Chain.
+     * validationID from the P-Chain. The validator weight change should not have any effect until this method is called.
      * @param messageIndex The index of the Warp message to be received providing the acknowledgement.
      */
     function completeValidatorWeightUpdate(
         uint32 messageIndex
-    ) external;
+    ) internal;
 }
 ```
 
-**Note:** The NodeIDs are stored as `bytes32` in the `ACP99Manager` interface to simplify the implementation.
+> **Note:** `IACP99Manager` does not include a method to return all active validations. This is because a `mapping` is a reasonable way to store active validations internally, and Solidity `mapping`s are not iterable. This can be worked around by storing additional indexing metadata in the contract, but not all applications may wish to incur that added complexity.
 
 #### About `Validation`s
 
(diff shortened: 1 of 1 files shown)


ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -226,150 +191,82 @@ interface IACP99Manager {
...
+The two-contract design consists of a contract that implements `ACP99Manager`, and a separate "security module" contract that implements a single staking security model, such as PoS or PoA. Each `ACP99Manager` implementation contract is associated with one "security module" that is the only contract allowed to call the `ACP99Manager` functions that initiate validator set changes (`initiateValidatorRegistration`, and `initiateValidatorWeightUpdate`). Every time a validator is added/removed or a weight change is initiated, the `ACP99Manager` implementation will in turn call the corresponding function of the "security module" (`handleValidatorRegistration` or `handleValidatorWeightChange`). We recommend that the "security module" reference an immutable `ACP99Manager` contract address for security reasons.
```
Nuttymoon, line 263: I wouldn't qualify PoA as "staking".
```suggestion
The multi-contract design consists of a contract that implements `ACP99Manager`, and separate "security module" contracts that implement security models, such as PoS or PoA. Each `ACP99Manager` implementation contract is associated with one or more "security modules" that are the only contracts allowed to call the `ACP99Manager` functions that initiate validator set changes (`initiateValidatorRegistration`, and `initiateValidatorWeightUpdate`). Every time a validator is added/removed or a weight change is initiated, the `ACP99Manager` implementation will, in turn, call the corresponding function of the "security module" (`handleValidatorRegistration` or `handleValidatorWeightChange`). We recommend that the "security modules" reference an immutable `ACP99Manager` contract address for security reasons.
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -226,150 +191,82 @@ interface IACP99Manager {
...
-     */
-    function handleValidatorWeightChange(
-        ValidatorWeightChangeInfo memory weightChangeInfo
-    ) external;
-}
-```
+`ACP99Manager` is designed to be easily incorporated into any architecture. Two example integrations are included in this ACP, each of which uses a different architecture.
 
-### Reference Architecture
+#### Two-contract Design
```
Nuttymoon, line 261: "Multi-contract Design" would be more generic (see the [BalancerValidatorManager](https://github.com/suzaku-network/suzaku-contracts-library/blob/main/src/contracts/ValidatorManager/BalancerValidatorManager.sol) that can have multiple security modules managing registrations/weight changes)
```suggestion
#### Multi-contract Design
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -226,150 +191,82 @@ interface IACP99Manager {
...
 
-It is up to the "security module" to decide what action to take when a validator is added/removed or a weight change is confirmed by the P-Chain. Such actions could be starting the withdrawal period and allocating rewards in a PoS Subnet.
+It is up to the "security module" to decide what action to take when a validator is added/removed or a weight change is confirmed by the P-Chain. Such actions could be starting the withdrawal period and allocating rewards in a PoS L1.
 
 ```mermaid
 ---
-title: ACP-99 Architecture with a PoA Security Module
+title: ACP-99 Two-Contract Architecture with a PoA Security Module
```
Nuttymoon, line 269:
```suggestion
title: ACP-99 Multi-contract Architecture with a PoA Security Module
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -394,36 +291,57 @@ graph LR
...
-If we don't keep track of the historical information in the `ACP99Manager` contract, this information will still be available in archive nodes and offchain tools (e.g. explorers).
-
-### Is an upgradeable architecture more appropriate?
-
-The Teleporter team has been working on its own implementation of a `ValidatorManager` contract in the [teleporter](https://github.com/ava-labs/teleporter/blob/main/contracts/validator-manager/ValidatorManager.sol) repository that is based on abstract and upgradeable contracts.
-
-### Should the `ACP99Manager` include a churn control mechanism?
+It's undefined if `getValidation` should return allow historical information about ended validations. Should `ACP99Manager` enforce that validation information be kept in the contract's state indefinitely? Note that validator performance (uptime) is _not_ specified in the `ACP99Manager` interface, as it may not be relevant to some applications (e.g. PoA). Historical uptime would be a useful metric to query, but it may be more appropriately left to implementations to enforce.
```
Nuttymoon, line 338:
```suggestion
It's undefined if `getValidation` should return historical information about ended validations. Should `ACP99Manager` enforce that validation information be kept in the contract's state indefinitely? Note that validator performance (uptime) is _not_ specified in the `ACP99Manager` interface, as it may not be relevant to some applications (e.g. PoA). Historical uptime would be a useful metric to query, but it may be more appropriately left to implementations to enforce.
```

Nuttymoon (changes requested):
Thanks a lot for the PR @cam-schultz!

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -58,160 +79,104 @@ struct PChainOwner {
...
- * @param activeSeconds The time during which the validator was active during this validation
- * @param uptimeSeconds The uptime of the validator for this validation
+ * @notice Contains the active state of a Validation
+ * @notice status is the validation status
+ * @notice nodeID is The NodeID of the validator
+ * @notice startingWeight is the weight of the validator at the time of registration
+ * @notice messageNonce is the current weight update nonce
+ * @notice weight is the current weight of the validator
+ * @notice startTime The start time of the validation
+ * @notice endTime The end time of the validation
```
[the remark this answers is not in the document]
cam-schultz: `@param` only pertains to [functions, events, and errors](https://docs.soliditylang.org/en/latest/natspec-format.html#tags), so I went with the more generic `@notice` here. Happy to change it to `@dev` if you think that would be more appropriate, but it seems NatSpec is not all that useful for documenting structs.

ACPs/99-validatorsetmanager-contract/README.md
cam-schultz: > Each `ACP99Manager` implementation contract is associated with one or more "security modules"

Curious if there is a specific use case in mind that would use more than one security module. Or are you clarifying that the security module need not be a single contract? Either way, implementation of such a design and enforcement by the `ValidatorManager` thereof is outside the scope of this ACP.

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -18,34 +18,55 @@ This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https:/
...
 
-Given these assumptions and the fact that most of the active blockchains on Avalanche mainnet are EVM-based, we propose defining a reference implementation for a Solidity smart contract that can:
+Given these assumptions and the fact that most of the active blockchains on Avalanche mainnet are EVM-based, we propose `ACP99Manager` as the standard Solidity contract specification that can:
 
-1. Hold relevant information about the current Subnet validator set, as well as historical information
+1. Hold relevant information about the current Subnet validator set
 2. Send validator set updates to the P-Chain by generating `AdressedCall`s defined in ACP-77
 3. Correctly update the validator set by interpreting notification messages received from the P-Chain
-4. Be easily extended with custom “security modules” to implement any security model (e.g. PoS). Those modules have to implement the `IACP99SecurityModule` interface and will be called by the `ACP99Manager` contract upon validator set updates.
+4. Be easily integrated into applications implementing various security models (e.g. Proof-of-Stake).
```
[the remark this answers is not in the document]
cam-schultz: Updated the language

ACPs/99-validatorsetmanager-contract/README.md
Nuttymoon: The use case we have in mind is allowing a smoother transition between security models (than what is available by having a single security module). See the [BalancerValidatorManager](https://github.com/suzaku-network/suzaku-contracts-library/blob/main/src/contracts/ValidatorManager/BalancerValidatorManager.sol) that allows to set a max total weight per security module. An example would be giving 800 max weight to a PoA security module and 200 max weight to a PoS security module so start decentralizing the network and then allocating more and more weight on the PoS side.

ACPs/99-validatorsetmanager-contract/README.md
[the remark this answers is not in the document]
Nuttymoon: Even tho the `@param` tag is officially supported only for functions, events, and errors, the documentation generation still looks better IMO. But I am okay with using `@notice` if you don't feel comfortable using `@param`.

Using `@notice`:
image

Using `@param`:
image

Nuttymoon, line 89: If sticking to using `@notice`:
```suggestion
 * @notice status is the validation status
 * @notice nodeID is the NodeID of the validator
 * @notice startingWeight is the weight of the validator at the time of registration
 * @notice messageNonce is the current weight update nonce
 * @notice weight is the current weight of the validator
 * @notice startTime is the start time of the validation
 * @notice endTime is the end time of the validation
```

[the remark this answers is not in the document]
cam-schultz: Nice, I hadn't seen it rendered and assumed @param wouldn't be supported in this context. Happy to switch over.

ACPs/99-validatorsetmanager-contract/README.md
cam-schultz: Very cool idea, thanks for linking.

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -58,160 +79,104 @@ struct PChainOwner {
...
-
-    /// @notice Get the address of the security module attached to this manager
-    function getSecurityModule() external view returns (address);
-
-    /// @notice Get the validation details for a given validation ID
-    function getValidation(
-        bytes32 validationID
-    ) external view returns (Validation memory);
+    /// @notice Returns the ID of the Subnet tied to this manager
+    function subnetID() virtual public view returns (bytes32);
```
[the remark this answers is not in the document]
cam-schultz: Technically, `subnetID` is correct here. It's not as explicit as it could be, but note that the ACP-77 message specifications use `subnetID` to identify the converted L1. Also note that the `subnetID` is the tx hash of the `CreateSubnetTx`, and therefore does not change when the Subnet is converted to an L1. As such, changing the name of the identifier of a static value depending on the Subnet/L1 status would be confusing in its own right. Eventually, once Subnets are fully deprecated in favor of L1s, it will make more sense to revisit this terminology.

This could be much better documented (and in fact, some contracts in the `icm-contracts` repo linked in this PR need to be changed back to `subnetID`), but ACP-77 is `Active` and therefore considered locked.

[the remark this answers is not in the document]
Nuttymoon: Makes sense. I only suggested changing here because v1.0.0 of `ValidatorManager` uses `_l1ID`:

https://github.com/ava-labs/icm-contracts/blob/046ca6d83c07c4cffecf61c06f4156b8b08725cc/contracts/validator-manager/ValidatorManager.sol#L40

Nuttymoon (approved):

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -18,34 +18,55 @@ This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https:/
 
 On each validator set change, the P-Chain is willing to sign an `AddressedCall` to notify any on-chain program tracking the validator set. On-chain programs must be able to interpret this message, so they can trigger the appropriate action. The 2 kinds of `AddressedCall`s [defined in ACP-77](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#p-chain-warp-message-payloads) are `L1ValidatorRegistrationMessage` and `L1ValidatorWeightMessage`.
 
-Given these assumptions and the fact that most of the active blockchains on Avalanche mainnet are EVM-based, we propose defining a reference implementation for a Solidity smart contract that can:
+Given these assumptions and the fact that most of the active blockchains on Avalanche mainnet are EVM-based, we propose `ACP99Manager` as the standard Solidity contract specification that can:
```
meaghanfitzgerald, line 21:
```suggestion
Given these assumptions and the fact that most of the active blockchains on Avalanche Mainnet are EVM-based, we propose `ACP99Manager` as the standard Solidity contract specification that can:
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -18,34 +18,55 @@ This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https:/
...
 
-1. Hold relevant information about the current Subnet validator set, as well as historical information
+1. Hold relevant information about the current Subnet validator set
```
meaghanfitzgerald, line 23:
```suggestion
1. Hold relevant information about the current L1 validator set
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -18,34 +18,55 @@ This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https:/
...
-4. Be easily extended with custom “security modules” to implement any security model (e.g. PoS). Those modules have to implement the `IACP99SecurityModule` interface and will be called by the `ACP99Manager` contract upon validator set updates.
+4. Be easily integrated into validator manager implementations that utilize various security models (e.g. Proof-of-Stake).
 
-Having an audited and open-source reference implementation freely available will contribute to lowering the cost of launching Subnets on Avalanche.
+Having an audited and open-source reference implementation freely available will contribute to lowering the cost of launching L1s on Avalanche.
 
-Once deployed, the `ACP99Manager` contract will be used as the `Address` in the [`ConvertSubnetToL1Tx`](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#convertsubnettol1tx).
+Once deployed, the `ACP99Manager` implementation contract will be used as the `Address` in the [`ConvertSubnetToL1Tx`](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#convertsubnettol1tx).
```
meaghanfitzgerald, line 30:
```suggestion
Once deployed, the `ACP99Manager` implementation contract can be used as the `Address` in the [`ConvertSubnetToL1Tx`](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#convertsubnettol1tx).
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -18,34 +18,55 @@ This ACP relies on concepts introduced in [ACP-77 (Reinventing Subnets)](https:/
```
meaghanfitzgerald, line 18: for some reason, I can't make suggestions on Line 17:
The P-Chain allows a Subnet to specify a "validator manager" if it chooses to convert to an L1 using `ConvertSubnetToL1Tx`. 
Outside of any direct reference to the `SubnetID` param, this ACP should only reference any management of L1 validators sets and Avalanche L1 networks.

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -58,160 +79,104 @@ struct PChainOwner {
...
-    uint64 activeSeconds;
-    uint64 uptimeSeconds;
-    uint256 activeWeightSeconds;
-    uint256 uptimeWeightSeconds;
-}
+#### About `Validation`s
+
+A `Validation` represents the continuous time frame during which a node is part of the validator set and can be composed of multiple periods. A new period starts every time the validator weight changes during the same validation.
+
+Each `Validation` is identified by its `validationID` which is the SHA256 of the Payload of the `AddressedCall` in the `RegisterL1ValidatorTx` adding the validator to the Subnet's validator set, as defined in ACP-77.
```
meaghanfitzgerald, line 106:
```suggestion
Each `Validation` is identified by its `validationID` which is the SHA256 of the Payload of the `AddressedCall` in the `RegisterL1ValidatorTx` adding the validator to the L1's validator set, as defined in ACP-77.
```

ACPs/99-validatorsetmanager-contract/README.md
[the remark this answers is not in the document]
meaghanfitzgerald: I'd suggest: 
```suggestion
    /// @notice Returns the SubnetID of the L1 tied to this manager
    function subnetID() virtual public view returns (bytes32);
```
or 
```suggestion
    /// @notice Returns the SubnetID of the L1 tied to this manager
    function l1ID() virtual public view returns (bytes32);
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -394,36 +291,57 @@ graph LR
 
 "Security modules" could implement PoS, Liquid PoS, etc. The specification of such smart contracts is out of the scope of this ACP.
 
-## Backwards Compatibility
+A work in progress implementation is available in the [Suzaku Contracts Library](https://github.com/suzaku-network/suzaku-contracts-library/blob/acp-99-implementation/README.md#acp99-contracts-library) repository. It will be updated until this ACP is considered `Implementable` based on the outcome of the discussion.
```
meaghanfitzgerald, line 294: @Nuttymoon this link to suzaku is still returning a 404

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -394,36 +291,57 @@ graph LR
...
-The Teleporter team implentation of the `ValidatorManager` contract includes a churn control mechanism that prevents too much weight from being added or removed from the validator set in a short amount of time.
+If we don't require `ACP99Manager` implementations to keep track of historical validations, this information will still be available in archive nodes and offchain tools (e.g. explorers).
 
-### How could we name “Security Modules”?
+### Should `ACP99Manager` include a churn control mechanism?
 
-I don’t really like this name but cannot come up with anything else.
+The Ava Labs [implementation](https://github.com/ava-labs/icm-contracts/blob/main/contracts/validator-manager/ValidatorManager.sol) of the `ValidatorManager` contract includes a churn control mechanism that prevents too much weight from being added or removed from the validator set in a short amount of time. Excessive churn can cause consensus failures, so it may be appropriate to require that churn tracking is implemented in some capacity.
 
 ## Acknowledgments
```
meaghanfitzgerald, line 347: @cam-schultz you definitely deserve a spot on this list!

ACPs/99-validatorsetmanager-contract/README.md
cam-schultz: Updated this language to clarify.

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -58,160 +79,104 @@ struct PChainOwner {
...
-    function subnetID() external view returns (bytes32);
-
-    /// @notice Get the address of the security module attached to this manager
-    function getSecurityModule() external view returns (address);
-
-    /// @notice Get the validation details for a given validation ID
-    function getValidation(
-        bytes32 validationID
-    ) external view returns (Validation memory);
+    /// @notice Returns the ID of the Subnet tied to this manager
```
meaghanfitzgerald, line 149:
```suggestion
    /// @notice Returns the SubnetID of the L1 tied to this manager
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -58,318 +79,198 @@ struct PChainOwner {
...
-    uint64 activeSeconds;
-    uint64 uptimeSeconds;
-    uint256 activeWeightSeconds;
-    uint256 uptimeWeightSeconds;
-}
+#### About `Validator`s
+
+A `Validator` represents the continuous time frame during which a node is part of the validator set and can be composed of multiple periods. A new period starts every time the validator's weight changes.
+
+Each `Validator` is identified by its `validationID` which is the SHA256 of the Payload of the `AddressedCall` in the `RegisterL1ValidatorTx` adding the validator to the L1's validator set, as defined in ACP-77.
```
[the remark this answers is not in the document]
meaghanfitzgerald:
```suggestion
Each `Validator` is identified by its `validationID`. If a validator was added as part of the initial set of continuous dynamic fee paying validators, its `validationID` is the SHA256 hash of the `ConvertSubnetToL1Tx` ID and the index of the initial validator. If a validator was added to the L1's validator set post-conversion, its `validationID` is the SHA256 of the payload of the `AddressedCall` in the `RegisterL1ValidatorTx` used to add it, as defined in ACP-77. 
```

cam-schultz:
FYI, I've made one signficant change to the spec since this PR was opened, that being separating the "Remove a Validator" and "Update a Validator's Weight" operations into their own distint sets of functions. Although both of these operations are performed with a `SetL1ValidatorWeightTx` on the P-Chain, reflecting that at the contract level made for some odd integration patterns. For example, a validator weight change is confirmed by verifying a `L1ValidatorWeightMessage` from the P-Chain. However, if a validator is removed, the validator is removed from the P-Chain's state so it will **not** be willing to sign that message type. Instead, it signs a `L1ValidatorRegistrationMessage` with `registered=false`. As such, `completeValidatorWeightUpdate` would require handling for both message types in a unified implementation, which is neither a minimal interface not ideal UX.

Separating these operations into `initiateValidatorWeightUpdate`/`completeValidatorWeightUpdate` and `initiateValidatorRemoval`/`completeValidatorRemoval` is a cleaner choice for the following reasons:
1. Simpler implementation - each set of methods handles exactly one operation
2. Improved UX for the same reason as point (1) above
3. Abstracts away what is essentially a P-Chain optimization

Tagging everyone who's reviews so far: @Nuttymoon @meaghanfitzgerald @michaelkaplan13

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -79,31 +79,33 @@ struct PChainOwner {
...
     uint64 startTime;
     uint64 endTime;
 }
 ```
 
-#### About `Validation`s
+#### About `Validator`s
 
-A `Validation` represents the continuous time frame during which a node is part of the validator set and can be composed of multiple periods. A new period starts every time the validator weight changes during the same validation.
+A `Validator` represents the continuous time frame during which a node is part of the validator set and can be composed of multiple periods. A new period starts every time the validator's weight changes.
```
Nuttymoon, line 106: Does it still make sense to talk about validation periods? (since they were removed from the spec)
```suggestion
A `Validator` represents the continuous time frame during which a node is part of the validator set.
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -115,136 +117,176 @@ For a full implementation, please see the [Reference Implementation](#reference-
...
  * @notice The ACP99Manager interface represents the functionality for sovereign L1
- * validator management, as specified in ACP-77
+ * validator management, as specified in ACP-77.
  */
 abstract contract ACP99Manager {
-    /// @notice Emitted when an initial validator is registered
+    /// @notice Emitted when an initial validator is registered.
     event RegisteredInitialValidator(
-        bytes32 indexed nodeID, bytes32 indexed validationID, uint64 weight
+        bytes32 indexed validationID, bytes nodeID, uint64 weight
```
Nuttymoon, line 125: What's the rationale for removing `nodeID` from the indexed topics? IMO it is valuable to filter those events by NodeID to find a node's validations history.

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -115,136 +117,176 @@ For a full implementation, please see the [Reference Implementation](#reference-
...
+     * @param messageIndex The index of the L1ValidatorRegistrationMessage to be received providing the acknowledgement.
+     * @return validationID The ID of the registered validator.
      */
     function completeValidatorRegistration(
         uint32 messageIndex
-    ) virtual public returns (bytes32);
+    ) virtual public returns  (bytes32 validationID);
+
+    /**
+    * @notice Initiates validator removal by issuing a L1ValidatorWeightMessage with the weight set to zero.
```
Nuttymoon, line 225:
```suggestion
    * @notice Initiates validator removal by issuing an L1ValidatorWeightMessage with the weight set to zero.
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -115,136 +117,176 @@ For a full implementation, please see the [Reference Implementation](#reference-
...
+    * Emits an {InitiatedValidatorRemoval} on success.
+    *
+    * @param validationID The ID of the validator to remove.
+    */
+    function _initiateValidatorRemoval(
+        bytes32 validationID
+    ) virtual internal;
+
+    /**
+    * @notice Completes validator removal by consuming an RegisterL1ValidatorMessage from the P-Chain acknowledging
```
Nuttymoon, line 237:
```suggestion
    * @notice Completes validator removal by consuming a RegisterL1ValidatorMessage from the P-Chain acknowledging
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -115,136 +117,176 @@ For a full implementation, please see the [Reference Implementation](#reference-
...
+    */
+    function completeValidatorRemoval(
+        uint32 messageIndex
+    ) virtual public returns (bytes32 validationID);
 
     /**
-     * @notice Initiate a validator weight update by issuing a SetL1ValidatorWeightTx Warp message.
-     * If the weight is 0, this initiates the removal of the validator from the L1. The validator weight change
-     * should not have any effect until completeValidatorWeightUpdate is successfully called.
+     * @notice Initiates validator weight update by issuing a L1ValidatorWeightMessage with a nonzero weight.
```
Nuttymoon, line 249:
```suggestion
     * @notice Initiates a validator weight update by issuing an L1ValidatorWeightMessage with a nonzero weight.
```

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -115,136 +117,176 @@ For a full implementation, please see the [Reference Implementation](#reference-
...
+    function _initiateValidatorWeightUpdate(
         bytes32 validationID,
         uint64 weight
-    ) virtual internal returns (uint64);
+    ) virtual internal returns (uint64 nonce, bytes32 messageID);
 
     /**
-     * @notice Completes the validator weight update process by returning an acknowledgement of the weight update of a
-     * validationID from the P-Chain. The validator weight change should not have any effect until this method is successfully called.
+     * @notice Completes the validator weight update process by consuming a L1ValidatorWeightMessage from the P-Chain
```
Nuttymoon, line 265:
```suggestion
     * @notice Completes the validator weight update process by consuming an L1ValidatorWeightMessage from the P-Chain
```

ACPs/99-validatorsetmanager-contract/README.md
meaghanfitzgerald: looks like it is still there, just swapped place and typing with `validationID`.

```
bytes32 indexed validationID, **bytes nodeID**, uint64 weight
```

Nuttymoon: Yes, but it is not indexed anymore.

meaghanfitzgerald: ah i understand now, and agree that it would be helpful for tracking metrics per node

ACPs/99-validatorsetmanager-contract/README.md
cam-schultz: Agreed, validation periods delineated by weight changes are out of scope.

ACPs/99-validatorsetmanager-contract/README.md
Nuttymoon: Makes sense. I was unaware of this limitation in Solidity.

In my first implementation of ACP-99, I used a `bytes32` to store the NodeID as it is capable of storing both the current 20-byte format and the ACP-20 32-byte format. Do you reckon that there would be other format > 32 bytes in the future?

ACPs/99-validatorsetmanager-contract/README.md
```diff
@@ -394,36 +326,57 @@ graph LR
...
-If we don't keep track of the historical information in the `ACP99Manager` contract, this information will still be available in archive nodes and offchain tools (e.g. explorers).
-
-### Is an upgradeable architecture more appropriate?
-
-The Teleporter team has been working on its own implementation of a `ValidatorManager` contract in the [teleporter](https://github.com/ava-labs/teleporter/blob/main/contracts/validator-manager/ValidatorManager.sol) repository that is based on abstract and upgradeable contracts.
-
-### Should the `ACP99Manager` include a churn control mechanism?
+It's undefined if `getValidator` should return historical information about historical validators. Should `ACP99Manager` enforce that validator information be kept in the contract's state indefinitely? Note that validator performance (uptime) is _not_ specified in the `ACP99Manager` interface, as it may not be relevant to some applications (e.g. PoA). Historical uptime would be a useful metric to query, but it may be more appropriately left to implementations to enforce.
```
richardpringle, line 373: Maybe "it's left to the implementor" instead of "it's undefined"?

richardpringle (approved):

geoff-vball (commented):
Just one nit, otherwise LGTM

ACPs/99-validatorsetmanager-contract/README.md
cam-schultz: I've rewritten this whole section to be less awkwardly worded.

ACPs/99-validatorsetmanager-contract/README.md
cam-schultz: After discussing with @geoff-vball, we agree that the best compromise is to accept `bytes nodeID` in functions, and emit `bytes20 nodeID` in events. Dynamic `bytes` in functions simplifies the upgrade path if the `nodeID` format ever changes. Since the implementation must convert from dynamic to fixed bytes to emit the event, we might as well emit `bytes20` to exactly reflect the current format. Future `nodeID` format changes would require contract validation changes anyway, so the event can be updated at that time.

Finally, since we do have to pay gas to convert `bytes` -> `bytes20` (around 800 gas in my tests), I updated the events to only emit the `nodeID` once per `validationID`. The `validationID` is indexed in every event, so the corresponding `nodeID` should be easily recoverable even if it's not emitted in the every event.

PR #175, opened 2025-01-29
[ACP-174] Dynamic EVM Gas Limits and Price Discovery Updates

michaelkaplan13:
Proposes that the C-Chain and Subnet-EVM chains adopt a dynamic fee mechanism similar to the one [introduced on the P-Chain as part of ACP-103](../103-dynamic-fees/README.md), with modifications to allow for block proposers (i.e. validators) to dynamically adjust the target gas consumption per unit time.


michaelkaplan13:
Closing to correct the branch name with proper ACP number.

PR #176, opened 2025-01-29
[ACP-176] Dynamic EVM Gas Limits and Price Discovery Updates

michaelkaplan13:
Proposes that the C-Chain and Subnet-EVM chains adopt a dynamic fee mechanism similar to the one [introduced on the P-Chain as part of ACP-103](../103-dynamic-fees/README.md), with modifications to allow for block proposers (i.e. validators) to dynamically adjust the target gas consumption per unit time.

[Rendered](https://github.com/michaelkaplan13/ACPs/blob/acp-176/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md)

### DIFF
--- a/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
+++ b/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
@@ -1,0 +1,138 @@
+| ACP | 176 |
+| :- | :- |
+| **Title** | Dynamic EVM Gas Limits and Price Discovery Updates |
+| **Author(s)** | Michael Kaplan ([@michaelkaplan13](https://github.com/michaelkaplan13)), Stephen Buttolph ([@StephenButtolph](https://github.com/StephenButtolph)) |
+| **Status** | Proposed ([Discussion](POPULATED BY MAINTAINER, DO NOT SET)) |
+| **Track** | Standards |
+
+## Abstract
+
+Proposes that the C-Chain and Subnet-EVM chains adopt a dynamic fee mechanism similar to the one [introduced on the P-Chain as part of ACP-103](../103-dynamic-fees/README.md), with modifications to allow for block proposers (i.e. validators) to dynamically adjust the target gas consumption per unit time.
+
+## Motivation
+
+Currently, the C-Chain has a static gas target of [15,000,000 gas](https://github.com/ava-labs/coreth/blob/39ec874505b42a44e452b8809a2cc6d09098e84e/params/avalanche_params.go#L32) per [10 second rolling window](https://github.com/ava-labs/coreth/blob/39ec874505b42a44e452b8809a2cc6d09098e84e/params/avalanche_params.go#L36), and uses a modified version of the [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) dynamic fee mechanism to adjust the base fee of blocks based on the gas consumed in the previously 10 second window. This has two notable drawbacks:
+
+1. The windower mechanism used to determine the base fee of blocks can lead to outsized spikes in the gas price when there is a large block. This is because after a large block that uses all of its gas limit, blocks that follow in the same window continue to result in increased gas prices even they are relatively small blocks that are under the target gas consumption since the time of the large time.
+2. The static gas target necessitates a required network upgrade in order to modify. This is cumbersome and makes it difficult for the network to adjust its capacity in response to performance optimization or hardware requirement increases.
+
+To better position Avalanche EVM chains, including the C-Chain, to be able to handle future increases in load, we propose replacing the above mechanism with one that better handles blocks that consume a large amount of gas, and that allows for validators to dynamically adjust the target rate of consumption.
+
+## Specification
+
+### Gas Price Determination
+
+The mechanism to determine the base fee of a block is the same as the one used in ACP-103 to determine the gas price of a block on the P-Chain. This mechanism calculates the gas price for a given block $b$ based on the following parameters:
+
+<div align="center">
+
+| | |
+|---|---|
+| $T$ | the target gas consumed per second |
+| $M$ | minimum gas price |
+| $K$ | gas price update constant |
+| $C$ | maximum gas capacity |
+| $R$ | gas capacity added per second |
+
+</div>
+
+The full specification of this mechanism can be found [here](../103-dynamic-fees/README.md).
+
+### Making $T$ Dynamic
+
+As noted above, the gas price determination mechanism relies on a target gas consumption per second, $T$, in order to calculate the gas price for a given block. $T$ will be adjusted dynamically according to the following specification.
+
+Prior to the activation of this mechanism, $T$ is initialized to a prior estimate of the network's desired target gas consumption per second, which must be a positive integer.
+
+After the execution of block $b$, $T$ can be adjusted up or down:
+
+$$\frac{-T}{D} \le targetChange \le \frac{T}{D}$$
+$$T = max(P, T + targetChange)$$
+
+Where:
+ - $D$ is the target gas consumption rate bound divisor
+ - $P$ is the global minimum allowed target gas consumption rate for the network, which must be at least $D$.
+ - $targetChange$ is set by the builder of block $b$ within the valid range. 
+ 
+ If $|targetChange| > \frac{T}{D}$, block $b$ is considered invalid. Block builders (i.e. validators), may set their desired value for $T$ (i.e. `desired_rate`) in their configuration, and $targetChange$ can then be calculated for blocks that they build according to:
+
+  ```python
+  # Calculates the amount by which to adjust the target gas consumed per second
+  def calc_gas_consumption_rate_change(current_rate: int, desired_rate: int) -> int:
+    max_change = current_rate / target_gas_consumption_rate_bound_divisor
+    if desired_rate > current_rate:
+        return min(desired_rate - current_rate, max_change)
+    else:
+        return -1 * min(current_rate - desired_rate, max_change)
+  ```
+
+As the value of $T$ dynamically adjusts after the execution of each block, the value of $R$ (capacity added per second) is also updated such that:
+
+$$R = T*2$$
+
+This ensures that the gas price remains equally reactive in either direction depending on how far the actual gas consumption rate is from the target, whether it is above or below.
+
+The value of $C$ must also adjust proportionately, so we set:
+
+$$C = R*10$$
+
+This means that the maximum stored gas capacity would be reached after 10 seconds where no blocks have been accepted.
+
+Note that the values of $T$, $R$, and $C$ are updated **after** the execution of block $b$, which means they only take effect in determining the gas price of block $b+1$. The change to each of these values in block $b$ does not effect the gas price for transaction includedin block $b$ itself.
+
+Allowing block builders to adjust the target gas consumption rate in blocks that they makes it such that the effect target gas consumption rate should loosely converge over time around the stake weighted average value set by validators of the network. This is because the number of blocks each validator produces is propotional to their stake weight. This means that an individual validator's effect on the resulting target gas consumption for the network is propotional to their stake weight.
+
+Currently, it is not valid for a block to contain zero transactions since it would not have any effect and would be a waste of resources to accept into the blockchain. However, with this change, it is possible for blocks with zero transactions to still effect value of $T$. To ensure that validators are able to help influence the value of $T$ even if there are no transactions to be included at the time they are proposing a block, blocks with no transactions that alter the value of $T$ will now be considered valid. This makes it such that validators would not need to create and include a no-op transaction just be able to produce a block to alter the current value of $T$ if there are no pending transactions when it is their turn to propose a block.
+
+### Configuration Parameters
+As noted above, the gas price determination mechanism depends on the values of $T$, $M$, $K$, $C$, and $R$ to be set as parameters. $T$ is adjusted dynamically from its initial value based on $D$ and $P$, and the values of $R$ and $C$ are derived from $T$. 
+
+Parameters at activation on the C-Chain are:
+
+<div align="center">
+
+| Parameter | Description | P-Chain Configuration|
+| - | - | - |
+| $T$ | initial target gas consumed per second | $1,500,000$ |
+| $M$ | minimum gas price | $1*10^{-18}$ AVAX  |
+| $K$ | gas price update constant | $2,164,043$ |
+| $D$ | target change bound divisor | $1024$ |
+| $P$ | minimum target gas consumption per second | $1,000,000$
+
+</div>
+
+$T$ was chosen as the current target gas consumption rate on the C-Chain. Avalanche L1s may change this value to match their current gas consumption rate if they would like.
+
+$M$ was chosen as the minimum possible denomination of the native EVM asset, such that the gas price will be more likely to consistently be in a range of price discovery. The price discovery mechanism has already been battle tested on the P-Chain (and prior to that on Ethereum for blob gas prices as defined by EIP-4844), giving confidence that it will correct react to any increase in network usage in order to prevent a DOS attack.
+
+$K$ was chosen such that at sustained maximum capacity ($T*2$ gas/second), the fee rate will double every ~30 seconds.
+
+$D$ was chosen to match the [gas limit bound divisor that Ethereum currently uses](https://github.com/ethereum/go-ethereum/blob/52766bedb9316cd6cddacbb282809e3bdfba143e/params/protocol_params.go#L26) to limit the amount that validators can change their execution layer gas limit.
+
+$P$ was chosen as a safe bound on the minimum target gas usage that is lower than the current target gas consumption rate on the C-Chain today. The target gas consumption rate will not move towards $P$ unless the majority of stake weight of the network specifies $P$ as their desired gas consumption rate target.
+
+## Backwards Compatibility
+
+The changes proposed in this ACP require a required network upgrade in order to take effect. Prior to its activation, the current gas limit and price discovery mechanisms will continue to be used. Its activation should have relatively minor compatibility effects on any developer tooling. Notably, transaction formats, and thus wallets, are not impacted. After its activation, given that the value of $C$ is dynamically adjusted, the maximum possible gas consumed by an individual block, and thus maximum possible consumed by an individual transaction, will also dynamically adjust. This can affect that maximum size of smart contract that can be deployed at any given time, but should not have significant practical impact.
+
+## Reference Implementation
+
+A full reference implementation has not been provided yet. It will be provided once this ACP is considered Implementable.
+
+## Security Considerations
+
+This ACP changes the mechanism for determining the gas price on Avalanche EVM chains. The gas price is meant to adopt dynamically to respond to changes in demand for using the chain. If it does not react as expected, the chain could be at risk for a DOS attack (if the usage price is too low), or over charge users during period of low activity. This price discovery mechanism has already been employed on the P-Chain, but should again be thouroughly tested for use on the C-Chain prior to activation on the Avalanche Mainnet.
+
+Further, this ACP also introduces a mechanism for validators to change the gas limit of the C-Chain. If this limit is set too high, it is possible that validator nodes will not be able to keep up in the processing of blocks. An upper bound on the maximum possible gas limit could be considered to try to mitigate this risk, though it would then take further required network upgrades to scale the network past that limit. 
+
+## Acknowledgments
+
+Thanks to the following non-exhaustive list of individuals for input, discussion, and feedback on this ACP.
+- [Emin Gün Sirer](https://x.com/el33th4xor)
+- [Luigi D'Onorio DeMeo](https://x.com/luigidemeo)
+- [Darioush Jalali](https://github.com/darioush)
+- [Aaron Buchwald](https://github.com/aaronbuchwald)
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,138 @@
...
+| $M$ | minimum gas price | $1*10^{-18}$ AVAX  |
+| $K$ | gas price update constant | $2,164,043$ |
+| $D$ | target change bound divisor | $1024$ |
+| $P$ | minimum target gas consumption per second | $1,000,000$
+
+</div>
+
+$T$ was chosen as the current target gas consumption rate on the C-Chain. Avalanche L1s may change this value to match their current gas consumption rate if they would like.
+
+$M$ was chosen as the minimum possible denomination of the native EVM asset, such that the gas price will be more likely to consistently be in a range of price discovery. The price discovery mechanism has already been battle tested on the P-Chain (and prior to that on Ethereum for blob gas prices as defined by EIP-4844), giving confidence that it will correct react to any increase in network usage in order to prevent a DOS attack.
```
geoff-vball, line 106:
```suggestion
$M$ was chosen as the minimum possible denomination of the native EVM asset, such that the gas price will be more likely to consistently be in a range of price discovery. The price discovery mechanism has already been battle tested on the P-Chain (and prior to that on Ethereum for blob gas prices as defined by EIP-4844), giving confidence that it will correctly react to any increase in network usage in order to prevent a DOS attack.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+$P$ was chosen as a safe bound on the minimum target gas usage that is lower than the current target gas consumption rate on the C-Chain today. The target gas consumption rate will not move towards $P$ unless the majority of stake weight of the network specifies $P$ as their desired gas consumption rate target.
+
+## Backwards Compatibility
+
+The changes proposed in this ACP require a required network upgrade in order to take effect. Prior to its activation, the current gas limit and price discovery mechanisms will continue to be used. Its activation should have relatively minor compatibility effects on any developer tooling. Notably, transaction formats, and thus wallets, are not impacted. After its activation, given that the value of $C$ is dynamically adjusted, the maximum possible gas consumed by an individual block, and thus maximum possible consumed by an individual transaction, will also dynamically adjust. This can affect that maximum size of smart contract that can be deployed at any given time, but should not have significant practical impact.
```
geoff-vball, line 116: The maximum size of smart contract that can be deployed at any given time should not have significant practical impact? 👀

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Reference Implementation
+
+A full reference implementation has not been provided yet. It will be provided once this ACP is considered Implementable.
+
+## Security Considerations
+
+This ACP changes the mechanism for determining the gas price on Avalanche EVM chains. The gas price is meant to adopt dynamically to respond to changes in demand for using the chain. If it does not react as expected, the chain could be at risk for a DOS attack (if the usage price is too low), or over charge users during period of low activity. This price discovery mechanism has already been employed on the P-Chain, but should again be thouroughly tested for use on the C-Chain prior to activation on the Avalanche Mainnet.
```
geoff-vball, line 124:
```suggestion
This ACP changes the mechanism for determining the gas price on Avalanche EVM chains. The gas price is meant to adapt dynamically to respond to changes in demand for using the chain. If it does not react as expected, the chain could be at risk for a DOS attack (if the usage price is too low), or over charge users during period of low activity. This price discovery mechanism has already been employed on the P-Chain, but should again be thoroughly tested for use on the C-Chain prior to activation on the Avalanche Mainnet.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+<div align="center">
+
+| | |
+|---|---|
+| $T$ | the target gas consumed per second |
+| $M$ | minimum gas price |
+| $K$ | gas price update constant |
+| $C$ | maximum gas capacity |
+| $R$ | gas capacity added per second |
```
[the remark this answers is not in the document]
michaelkaplan13: `D` and `P` aren't defined/used in ACP-103 to calculate the gas price, so I think it makes more sense to introduce them and define them below where they are needed.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
michaelkaplan13: Ah, worth noting that the static contract size limit is still in place. It's just that some contract deployments take significant gas usage, primarily when they include complex constructor/initialization logic. So nothing specific to contracts necessarily, it's just that the maximum gas that a single tx can use will fluctuate, and contract deployments can tend to use significant gas.

michaelkaplan13: I'll reword to clarify

geoff-vball: Ahh okay :)

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+Parameters at activation on the C-Chain are:
+
+<div align="center">
+
+| Parameter | Description | P-Chain Configuration|
+| - | - | - |
+| $T$ | initial target gas consumed per second | $1,500,000$ |
+| $M$ | minimum gas price | $1*10^{-18}$ AVAX  |
+| $K$ | gas price update constant | $2,164,043$ |
```
StephenButtolph, line 102: If we want to keep the time for price to double a constant, then I think `K` needs to be set proportionally as well.

Specifically, if the time to double is `30s`, then:

$K = \dfrac{30\text{s}}{\ln(2)} \cdot \left(R - T\right) = \dfrac{30\text{s}}{\ln(2)} \cdot T$

Which means:
```suggestion
| $K$ | gas price update constant | $64,921,277$ |
```

We may want to simplify with:

$\dfrac{30\text{s}}{\ln(2)} \approx 43\text{s}$

Which would mean:

$K = 43 \cdot T$

Which would initially set:

```suggestion
| $K$ | gas price update constant | $64,500,000$ |
```

StephenButtolph: We will also need to modify `x` when we are modifying `K` so that the rate of change of the gas price is not dependent on `D`.

Specifically, when modifying $K_0$ -> $K_1$ we also need to modify $x_0$ to $x_1$ where $x_1 = \dfrac{x_0 \cdot K_1}{K_0}$

StephenButtolph: I also wonder if we should be utilizing a better mechanism for modifying $T$.

Specifically, lets say the community is 50/50 split. One side wants the target to go to $0$, the other side wants the target to go to $\infty$.

$\dfrac{1025}{1024} \cdot \dfrac{1023}{1024} = \dfrac{1048575}{1048576}$ which means the gas target would actually go down over time (this is why EIP-1559 ends up averaging under the target usage).

We could instead utilize the 4844 mechanism again, where $T = P \cdot e^\left(\frac{q}{D}\right)$ and then the block proposer can choose to increase/decrease `q` within some bounds.

michaelkaplan13: Interesting, yeah that makes sense to me. If we want to keep roughly the same bounds on how much `T` can change per block:

$$ T_{n+1} \leq T{n} \cdot \frac{D+1}{D}$$
$$...$$
$$e^{\frac{q_{n+1}}{D}} \leq e^{\frac{q_{n}}{D}} \cdot \frac{D+1}{D} $$
$$...$$
$$\frac{q_{n+1}}{D} \leq \frac{q_{n}}{D} + ln(\frac{D+1}{D}) $$
$$...$$
$$\Delta{q} \leq ln(\frac{D+1}{D}) \cdot D$$

`ln((d+1)/d)*d` approaches 1 as `d->infinity`. Assuming we want to keep each block as being allowed to changed `T` by roughly 1/1024 of it's current value, we could keep `D=1024` and allow `q` to change +/- 1 each block, or to allow for more granularity, set `D=8192` and allow `q` to change at most +/- 8 each block.

michaelkaplan13: Was updating this in the spec, but then thought that we may as well remove `D` from the price calculation and let validators set any values for `n` and `d`

$$T = P \cdot e^{n/d}$$

as long as

$$\frac{D-1}{D} \cdot \frac{n_{0}}{d_{0}} \leq \frac{n_{1}}{d_{1}} \leq \frac{D+1}{D} \cdot \frac{n_{0}}{d_{0}}$$


Thoughts?

StephenButtolph: So, I'm not sure if I like this tbh... But I don't think we should be including any $\dfrac{D \pm 1}{D}$, as I think that introduces the EIP-1559 weirdness. I think as far as the math is concerned... If we start with:

$T_0 = P \cdot e^\dfrac{n_0}{d_0}$

We want to ensure that:

$T_1 = P \cdot e^\dfrac{n_1}{d_1}$ is valid. The most that $n$ should move is $\frac{d_0}{D}$ so:

$P \cdot e^\dfrac{max\left(n_0 - \frac{d_0}{D}, 0\right)}{d_0} \leq P \cdot e^\dfrac{n_1}{d_1} \leq P \cdot e^\dfrac{n_0 + \frac{d_0}{D}}{d_0}$

Divide by $P$:

$e^\dfrac{max\left(n_0 - \frac{d_0}{D}, 0\right)}{d_0} \leq e^\dfrac{n_1}{d_1} \leq e^\dfrac{n_0 + \frac{d_0}{D}}{d_0}$

Log everything:

$\dfrac{max\left(n_0 - \frac{d_0}{D}, 0\right)}{d_0} \leq \dfrac{n_1}{d_1} \leq \dfrac{n_0 + \frac{d_0}{D}}{d_0}$

Multiply by $d_0$:

$max\left(n_0 - \frac{d_0}{D}, 0\right) \leq \dfrac{n_1 \cdot d_0}{d_1} \leq n_0 + \frac{d_0}{D}$

Multiply by $d_1$:

$max\left(n_0 \cdot d_1 - \frac{d_0 \cdot d_1}{D}, 0\right) \leq n_1 \cdot d_0 \leq n_0 \cdot d_1 + \frac{d_0 \cdot d_1}{D}$

Multiply by $D$:

$max\left(n_0 \cdot d_1 \cdot D - d_0 \cdot d_1, 0\right) \leq n_1 \cdot d_0 \cdot D \leq n_0 \cdot d_1 \cdot D + d_0 \cdot d_1$

StephenButtolph: It feels to me like this is a bit over-specified... I feel like we could just set:

$T = P \cdot e^\frac{n}{2^{20}}$ where the block builder can modify $\left|\Delta n\right| \leq 2^{10}$

I don't think we need to give the block builders so much freedom on how to change the target gas... Also, only having one variable to choose from should make selecting the value for $n$ a lot easier if the block builder is targeting a specific $T$.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+Allowing block builders to adjust the target gas consumption rate in blocks that they produce makes it such that the effect target gas consumption rate should loosely converge over time around the stake-weighted average value set by validators of the network. This is because the number of blocks each validator produces is proportional to their stake weight. This means that an individual validator's effect on the resulting target gas consumption for the network is proportional to their stake weight.
+
+As noted in ACP-103, the maximum gas consumed in a given period of time $\Delta{t}$, is $r + R \cdot \Delta{t}$, where $r$ is the remaining gas capacity at the end of previous block execution. The upper bound across all $\Delta{t}$ is $C + R \cdot \Delta{t}$. Phrased different, the maximum amount of gas that can be consumed by any given block $b$ is:
+
+$$gasLimt_{b} = min(r + R \cdot \Delta{t}, C)$$
```
meaghanfitzgerald, line 87:
```suggestion
$$gasLimit_{b} = min(r + R \cdot \Delta{t}, C)$$ 
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
michaelkaplan13: 🤦 fell right back into the 1559 trap.

Yeah that makes sense to me. And I agree if we set $D$ sufficiently high, validators don't lose any significant granularity in their ability to chose their desired value for $T$.

I also added a blurb around how block builders can estimate their $q_{desired}$ based their configured target gas consumption rate. I think it should be safe for builders to use `math.Log` and round the result to the nearest integer, but confirmation there would be great.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+$$\frac{-T}{D} \le targetChange \le \frac{T}{D}$$
+$$T = max(P, T + targetChange)$$
+
+Where:
+ - $D$ is the target gas consumption rate bound divisor
+ - $P$ is the global minimum allowed target gas consumption rate for the network, which must be at least $D$.
+ - $targetChange$ is set by the builder of block $b$ within the valid range. 
+ 
+ If $|targetChange| > \frac{T}{D}$, block $b$ is considered invalid. Block builders (i.e. validators), may set their desired value for $T$ (i.e. `desired_rate`) in their configuration, and $targetChange$ can then be calculated for blocks that they build according to:
```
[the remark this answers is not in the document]
michaelkaplan13: Good call. I added a "Choosing a desired value for T" section now.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+As noted above, the gas price determination mechanism relies on a target gas consumption per second, $T$, in order to calculate the gas price for a given block. $T$ will be adjusted dynamically according to the following specification.
+
+Let $q$ be a non-negative integer that is initialized to 0 upon activation of this mechanism. Let the target gas consumption per second be expressed as:
+
+$$T = P \cdot e^{q/D}$$
+
+where $P$ is the global minimum allowed target gas consumption rate for the network, and $D$ is a constant that helps control the rate of change of the target gas consumption.
+
+After the execution of transactions in block $b$, the values $q$ can be increased or decreased up to $Q$. It must be the case that $q_{b-1} - Q \leq q_{b} \leq q_{b-1} + Q$, or block $b$ is considered invalid. The amount by which $q$ changes after executing block $b$ is specified by the block builder.
```
meaghanfitzgerald, line 51:
```suggestion
After the execution of transactions in block $b-1$, the values $q$ can be increased or decreased up to $Q$, or block $b$ will be considered invalid. The amount by which $q$ changes after executing block $b$ is specified by the block builder.
```
Both sentences are saying the same thing, and I think the first sentence is the most clear.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+$$C = R \cdot 10$$
+
+This means that the maximum stored gas capacity would be reached after 10 seconds where no blocks have been accepted.
+
+In order to keep roughly constant the time it takes for the gas price to double at sustained maximum network capacity usage, the value of $K$ used in the gas price determination mechanism must be updated proportionatly to $T$. When updating $T$ from $T_{n}$ to $T_{n+1}$, we update $K$ such that:
+
+$$K_{n+1} = K_{n} \cdot \frac{T_{n+1}}{T_{n}}$$
+
+In order to have the gas price not be directly impacted by the change in $K$, we also update $x$ proportionatly. When updating $x$ after executing a block, instead of setting $x = x + G$ as specified in ACP-103, we set:
```
meaghanfitzgerald, line 86:
```suggestion
In order to have the gas price not be directly impacted by the change in $K$, we also update $x$ (excess gas consumption) proportionally. When updating $x$ after executing a block, instead of setting $x = x + G$ as specified in ACP-103, we set:
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+$$R = T \cdot 2$$
+
+This ensures that the gas price remains equally reactive in either direction depending on how far the actual gas consumption rate is from the target, whether it is above or below.
+
+The value of $C$ must also adjust proportionately, so we set:
+
+$$C = R \cdot 10$$
+
+This means that the maximum stored gas capacity would be reached after 10 seconds where no blocks have been accepted.
```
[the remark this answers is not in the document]
michaelkaplan13: $C$ is the maximum stored capacity. The 10 seconds part of it is just a factor of $C = R \cdot 10$, since $R$ is the capacity added per second.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,157 @@
...
+  def calc_next_q(q_current: int, q_desired: int, max_change: int) -> int:
+    if q_desired > q_current:
+        return q_current + min(q_desired - q_current, max_change)
+    else:
+        return q_current - min(q_current - q_desired, max_change)
+  ```
+
+As $q$ is updated after the execution of transactions within the block, $T$ is also updated such that $T = P \cdot e^{q/D}$ at all times. As the value of $T$ adjusts, the value of $R$ (capacity added per second) is also updated such that:
+
+$$R = T \cdot 2$$
```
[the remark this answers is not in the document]
michaelkaplan13: I tried to explain in the sentence below, but `R = 2T` makes it such that being `n` over the target and `n` below result in the same absolute value of change in the gas price. 

For instance, if this weren't the case, then being 1,000 above the target consumption rate could increase the price more than being 1,000 below the target consumption rate would decrease it.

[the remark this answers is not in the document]
michaelkaplan13: This is why ACP-103 parameterized `R = 2T`, but now that `T` is being made dynamic `R` must also be dynamic to have this property hold.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
michaelkaplan13: I don't think these are the same actually. The builder of block $b$ gets to specify the change in $q$ that takes effect after the transactions in the block are executed, meaning it effects the gas price of transactions starting in block $b+1$. I added the extra specification just to make the block validation rule crystal clear.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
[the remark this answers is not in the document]
michaelkaplan13: Sorry I'm not quite sure what you mean. Based solely on its definition, it is not the cumulative sum of R over 10 seconds, it's the maximum amount of capacity that can be stored for future use. Here I'm just saying that the value of $C$ will adapt to always be 10x R, such that the maximum stored capacity is always reached after 10 seconds of no gas usage (if starting from 0), no matter what the value of R is.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+| **Status** | Proposed ([Discussion](POPULATED BY MAINTAINER, DO NOT SET)) |
+| **Track** | Standards |
+
+## Abstract
+
+Proposes that the C-Chain and Subnet-EVM chains adopt a dynamic fee mechanism similar to the one [introduced on the P-Chain as part of ACP-103](../103-dynamic-fees/README.md), with modifications to allow for block proposers (i.e. validators) to dynamically adjust the target gas consumption per unit time.
+
+## Motivation
+
+Currently, the C-Chain has a static gas target of [15,000,000 gas](https://github.com/ava-labs/coreth/blob/39ec874505b42a44e452b8809a2cc6d09098e84e/params/avalanche_params.go#L32) per [10 second rolling window](https://github.com/ava-labs/coreth/blob/39ec874505b42a44e452b8809a2cc6d09098e84e/params/avalanche_params.go#L36), and uses a modified version of the [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) dynamic fee mechanism to adjust the base fee of blocks based on the gas consumed in the previously 10 second window. This has two notable drawbacks:
```
StephenButtolph, line 14:
```suggestion
Currently, the C-Chain has a static gas target of [15,000,000 gas](https://github.com/ava-labs/coreth/blob/39ec874505b42a44e452b8809a2cc6d09098e84e/params/avalanche_params.go#L32) per [10 second rolling window](https://github.com/ava-labs/coreth/blob/39ec874505b42a44e452b8809a2cc6d09098e84e/params/avalanche_params.go#L36), and uses a modified version of the [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) dynamic fee mechanism to adjust the base fee of blocks based on the gas consumed in the previous 10 second window. This has two notable drawbacks:
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+1. The windower mechanism used to determine the base fee of blocks can lead to outsized spikes in the gas price when there is a large block. This is because after a large block that uses all of its gas limit, blocks that follow in the same window continue to result in increased gas prices even if they are relatively small blocks that are under the target gas consumption.
+2. The static gas target necessitates a required network upgrade in order to modify. This is cumbersome and makes it difficult for the network to adjust its capacity in response to performance optimization or hardware requirement increases.
```
StephenButtolph, line 17:
```suggestion
2. The static gas target necessitates a required network upgrade in order to modify. This is cumbersome and makes it difficult for the network to adjust its capacity in response to performance optimizations or hardware requirement increases.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+As noted above, the gas price determination mechanism relies on a target gas consumption per second, $T$, in order to calculate the gas price for a given block. $T$ will be adjusted dynamically according to the following specification.
+
+Let $q$ be a non-negative integer that is initialized to 0 upon activation of this mechanism. Let the target gas consumption per second be expressed as:
+
+$$T = P \cdot e^{q/D}$$
+
+where $P$ is the global minimum allowed target gas consumption rate for the network, and $D$ is a constant that helps control the rate of change of the target gas consumption.
+
+After the execution of transactions in block $b$, the value of $q$ can be increased or decreased up to $Q$. It must be the case that $q_{b-1} - Q \leq q_{b} \leq q_{b-1} + Q$, or block $b$ is considered invalid. The amount by which $q$ changes after executing block $b$ is specified by the block builder.
```
StephenButtolph, line 51:
```suggestion
After the execution of transactions in block $b$, the value of $q$ can be increased or decreased up to $Q$. It must be the case that $|\Delta q| \leq Q$, or block $b$ is considered invalid. The amount by which $q$ changes after executing block $b$ is specified by the block builder.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+The full specification of this mechanism can be found [here](../103-dynamic-fees/README.md).
+
+### Making $T$ Dynamic
+
+As noted above, the gas price determination mechanism relies on a target gas consumption per second, $T$, in order to calculate the gas price for a given block. $T$ will be adjusted dynamically according to the following specification.
+
+Let $q$ be a non-negative integer that is initialized to 0 upon activation of this mechanism. Let the target gas consumption per second be expressed as:
+
+$$T = P \cdot e^{q/D}$$
```
StephenButtolph, line 47:
```suggestion
$$T = P \cdot e^{\frac{q}{D}}$$
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+where $P$ is the global minimum allowed target gas consumption rate for the network, and $D$ is a constant that helps control the rate of change of the target gas consumption.
+
+After the execution of transactions in block $b$, the value of $q$ can be increased or decreased up to $Q$. It must be the case that $q_{b-1} - Q \leq q_{b} \leq q_{b-1} + Q$, or block $b$ is considered invalid. The amount by which $q$ changes after executing block $b$ is specified by the block builder.
+
+Block builders (i.e. validators), may set their desired value for $T$ (i.e. their desired gas consumption rate) in their configuration, and their desired value for $q$ can then be calculated as:
+
+$$q_{desired} = D \cdot ln(\frac{T}{P})$$
```
StephenButtolph, line 55:
```suggestion
$$q_{desired} = D \cdot ln\left(\frac{T}{P}\right)$$
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+Note that since $q_{desired}$ is only used locally and can be different for each node, it is safe for implementations to approximate the value of $ln(\frac{T}{P})$, and round the resulting value to the nearest integer.
```
StephenButtolph, line 57:
```suggestion
Note that since $q_{desired}$ is only used locally and can be different for each node, it is safe for implementations to approximate the value of $ln\left(\frac{T}{P}\right)$, and round the resulting value to the nearest integer.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+When building a block, builders can calculate their next preferred value for $q$ according to:
+
+  ```python
+  # Calculates a node's new desired value for q given for a given block
+  def calc_next_q(q_current: int, q_desired: int, max_change: int) -> int:
+    if q_desired > q_current:
+        return q_current + min(q_desired - q_current, max_change)
+    else:
+        return q_current - min(q_current - q_desired, max_change)
+  ```
```
StephenButtolph, line 68: I spent so long looking at this to see if it could be simplified... There probably is something simpler... but I couldn't figure out how haha

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+As $q$ is updated after the execution of transactions within the block, $T$ is also updated such that $T = P \cdot e^{q/D}$ at all times. As the value of $T$ adjusts, the value of $R$ (capacity added per second) is also updated such that:
+
+$$R = T \cdot 2$$
+
+This ensures that the gas price remains equally reactive in either direction depending on how far the actual gas consumption rate is from the target, whether it is above or below.
```
StephenButtolph, line 74:
```suggestion
This ensures that the gas price can increase and decrease at the same rate.
```

Specifically, the rate of gas consumption is in `[0, C]`. Because $C = 2T$, then excess will be increasing at a rate of `T` when the gas consumption rate is `C` and decreasing at a rate of `T` If the gas consumption rate is `0`.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+The value of $C$ must also adjust proportionately, so we set:
+
+$$C = R \cdot 10$$
+
+This means that the maximum stored gas capacity would be reached after 10 seconds where no blocks have been accepted.
+
+In order to keep roughly constant the time it takes for the gas price to double at sustained maximum network capacity usage, the value of $K$ used in the gas price determination mechanism must be updated proportionatly to $T$. When updating $T$ from $T_{n}$ to $T_{n+1}$, we update $K$ such that:
+
+$$K_{n+1} = K_{n} \cdot \frac{T_{n+1}}{T_{n}}$$
```
StephenButtolph, line 84:
```suggestion
In order to keep roughly constant the time it takes for the gas price to double at sustained maximum network capacity usage, the value of $K$ used in the gas price determination mechanism must be updated proportionally to $T$ such that:

$$K = T \cdot 43$$
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+In order to have the gas price not be directly impacted by the change in $K$, we also update $x$ (excess gas consumption) proportionally. When updating $x$ after executing a block, instead of setting $x = x + G$ as specified in ACP-103, we set:
+
+$$x_{n+1} = (x + G) \cdot \frac{K_{n+1}}{K_{n}}$$
+
+Note that the value of $q$ (and thus also $T$, $R$, $C$, $K$, and $x$) are updated **after** the execution of block $b$, which means they only take effect in determining the gas price of block $b+1$. The change to each of these values in block $b$ does not effect the gas price for transactions included in block $b$ itself.
+
+Allowing block builders to adjust the target gas consumption rate in blocks that they produce makes it such that the effect target gas consumption rate should converge over time around the stake-weighted average value set by validators of the network. This is because the number of blocks each validator produces is proportional to their stake weight. This means that an individual validator's effect on the resulting target gas consumption for the network is proportional to their stake weight.
```
StephenButtolph, line 92: > the effect target gas consumption rate should converge over time around the stake-weighted average value set by validators of the network.

I don't think this is quite accurate. I think the effective target gas consumption rate should converge to the point where 50% of voting stake wants it decreased and 50% of voting stake wants it increased.

If 49% of the network wanted the target gas consumption to go to 0 and 51% of the network wanted the target gas consumption to go to $x$, then the target would move towards $x$ (albeit slowly) over time.

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+As noted in ACP-103, the maximum gas consumed in a given period of time $\Delta{t}$, is $r + R \cdot \Delta{t}$, where $r$ is the remaining gas capacity at the end of previous block execution. The upper bound across all $\Delta{t}$ is $C + R \cdot \Delta{t}$. Phrased different, the maximum amount of gas that can be consumed by any given block $b$ is:
```
StephenButtolph, line 94:
```suggestion
As noted in ACP-103, the maximum gas consumed in a given period of time $\Delta{t}$, is $r + R \cdot \Delta{t}$, where $r$ is the remaining gas capacity at the end of previous block execution. The upper bound across all $\Delta{t}$ is $C + R \cdot \Delta{t}$. Phrased differently, the maximum amount of gas that can be consumed by any given block $b$ is:
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Configuration Parameters
+
+As noted above, the gas price determination mechanism depends on the values of $T$, $M$, $K$, $C$, and $R$ to be set as parameters. $T$ is adjusted dynamically from its initial value based on $D$ and $P$, and the values of $R$ and $C$ are derived from $T$. 
+
+Parameters at activation on the C-Chain are:
+
+<div align="center">
+
+| Parameter | Description | P-Chain Configuration|
```
StephenButtolph, line 108:
```suggestion
| Parameter | Description | C-Chain Configuration|
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+| $D$ | target gas consumption rate update constant | $2^{25}$ |
+| $Q$ | target gas consumption rate update factor change limit | $2^{15}$ |
+| $M$ | minimum gas price | $1*10^{-18}$ AVAX  |
+| $K$ | initial gas price update factor | $64,500,000$ |
+
+</div>
+
+$P$ was chosen as a safe bound on the minimum target gas usage on the C-Chain. The current gas target of the C-Chain is $1,500,000$ per second. The target gas consumption rate will only stay at $P$ if the majority of stake weight of the network specifies $P$ as their desired gas consumption rate target. 
+
+$D$ and $Q$ were chosen to give each block builder the ability to adjust the value of $T$ by roughly $\frac{1}{1024}$ of its current value, which matches the [gas limit bound divisor that Ethereum currently uses](https://github.com/ethereum/go-ethereum/blob/52766bedb9316cd6cddacbb282809e3bdfba143e/params/protocol_params.go#L26) to limit the amount that validators can change the execution layer gas limit in a single block. $D$ and $Q$ were scaled up by a factor of $2^{15}$ to block builders more granularity in the adjustments to $T$ that they would like to make.
```
StephenButtolph, line 120:
```suggestion
$D$ and $Q$ were chosen to give each block builder the ability to adjust the value of $T$ by roughly $\frac{1}{1024}$ of its current value, which matches the [gas limit bound divisor that Ethereum currently uses](https://github.com/ethereum/go-ethereum/blob/52766bedb9316cd6cddacbb282809e3bdfba143e/params/protocol_params.go#L26) to limit the amount that validators can change the execution layer gas limit in a single block. $D$ and $Q$ were scaled up by a factor of $2^{15}$ to provide block builders more granularity in the adjustments to $T$ that they can make.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+$K$ was chosen such that at sustained maximum capacity ($T*2$ gas/second), the fee rate will double every ~29.8 seconds.
+
+### Choosing a desired value for $T$
+
+As mentioned above, this new mechanism allows for validators to specify their desired target gas consumption rate ($T$) in their configuration, and the value that they set impacts the effective target gas consumption rate of the network over time. The higher the value of $T$, the more resources (storage, compute, etc) that are able to be used by the network. When choosing what value makes sense for them, validators should consider the resources that are required to properly support that level of gas consumption, the utility the network provides by having higher transaction per second throughput, and the stability of network should it reach that level of utilization.
+
+While Avalanche Network Clients can set default configuration values for the desired target gas consumption rate, each validator can choose to set this value independent based on their own considerations.
```
StephenButtolph, line 130:
```suggestion
While Avalanche Network Clients can set default configuration values for the desired target gas consumption rate, each validator can choose to set this value independently based on their own considerations.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Backwards Compatibility
+
+The changes proposed in this ACP require a required network upgrade in order to take effect. Prior to its activation, the current gas limit and price discovery mechanisms will continue to be used. Its activation should have relatively minor compatibility effects on any developer tooling. Notably, transaction formats, and thus wallets, are not impacted. After its activation, given that the value of $C$ is dynamically adjusted, the maximum possible gas consumed by an individual block, and thus maximum possible consumed by an individual transaction, will also dynamically adjust. The upper bound on the amount of gas consumed by a single transaction fluctuating means that transactions that are considered invalid and one time may be considered valid at a different point in time, and vice versa. While potentially unintuitive, as long as the minimum gas consumption rate is set sufficiently high this should not have significant practical impact, and is also currently the case on the Ethereum mainnet.
```
StephenButtolph, line 134:
```suggestion
The changes proposed in this ACP require a required network upgrade in order to take effect. Prior to its activation, the current gas limit and price discovery mechanisms will continue to be used. Its activation should have relatively minor compatibility effects on any developer tooling. Notably, transaction formats, and thus wallets, are not impacted. After its activation, given that the value of $C$ is dynamically adjusted, the maximum possible gas consumed by an individual block, and thus maximum possible consumed by an individual transaction, will also dynamically adjust. The upper bound on the amount of gas consumed by a single transaction fluctuating means that transactions that are considered invalid at one time may be considered valid at a different point in time, and vice versa. While potentially unintuitive, as long as the minimum gas consumption rate is set sufficiently high this should not have significant practical impact, and is also currently the case on the Ethereum mainnet.
```

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
michaelkaplan13: The 43 will probably seems like a magic number here for someone reading it through the first time, but it is a simpler specification 👍

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
michaelkaplan13: I think correctness is really the only thing that matters here anyways, unless you're thinking of implementing a python client in the near future 😄

ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
michaelkaplan13: Interesting, I had not realized that...so a stake weight equilibrium rather than average in a way.

michaelkaplan13: I updated it now, and removed the final sentence for conciseness.

StephenButtolph (approved):

meaghanfitzgerald (approved):

PR #181, opened 2025-02-08
ACP-181 ProposerVM Epochs

cam-schultz:
Proposes enabling epochs in the ProposerVM to provide more stable views of the P-Chain to inner VMs

[Rendered](https://github.com/cam-schultz/ACPs/blob/3b790e45096a331eb356dede9fde868444a924b7/ACPs/181-proposervm-epochs/README.md)

### DIFF
--- a/ACPs/181-proposervm-epochs/README.md
+++ b/ACPs/181-proposervm-epochs/README.md
@@ -1,0 +1,82 @@
+| ACP | 181 |
+| :--- | :--- |
+| **Title** | ProposerVM Epochs |
+| **Author(s)** | Cam Schultz [@cam-schultz](https://github.com/cam-schultz) |
+| **Status** | Proposed, Implementable, Activated, Stale ([Discussion](POPULATED BY MAINTAINER, DO NOT SET)) |
+| **Track** | Standards |
+
+## Abstract
+
+Proposes the introduction of epochs to the ProposerVM, such that a consistent view of the P-Chain may be provided to inner VMs for a known duration of time. This would enable VMs to optimize validator set retrievals, which currently must be done as often as every P-Chain block.
+
+## Motivation
+
+The P-Chain maintains a registry of L1 and Subnet validators (including Primary Network validators). Validators are added, removed, or their weights changed by issuing P-Chain transactions that are included in P-Chain blocks. When describing an L1 or Subnet's validator set, what is really being described are the weights, BLS keys, and Node IDs of the active validators at a particular P-Chain height. Use cases that require on-demand views of L1 or Subnet validator sets need to fetch validator sets at arbitrary P-Chain heights, while use cases that require up-to-date views need to fetch them as often as every P-Chain block.
+
+ProposerVM epochs during which the P-Chain height is fixed would widen this window to a predictable epoch duration, allowing these use cases to implement optimizations such as pre-fetching validator sets once per epoch, or allowing more efficient backwards traversal of the P-Chain to fetch historical validator sets.
+
+## Specification
+
+### Epoch Definition
+
+An epoch $E_n$ is defined by its start time $T_{start}^n$ and its end time $T_{end}^n$. Any block with timestamp $t$ is included in epoch $E_n$ if $T_{start}^n <= t < T_{end}^n$. The first block whose timestamp is greater than or equal to $T_{end}^n$ and whose *parent's* timestamp is less than $t_{start}^n$ will be the final block in $E_n$. Formally, for a block with timestamp $t$ and parent timestamp $t_{parent}, the block *seals* $E_n$ if 
+
+$$
+t >= T_{end}^n \: \text{ and } \: t_{parent} < T_{start}^n
+$$
+
+The next block after the block that seals $E_n$ is the first block of epoch $E_{n+1}$.
+
+### Epoch Duration
+
+Epochs have a constant duration $D$, such that for any given epoch $E_n$, the difference between its start and end times equals this duration: 
+
+$$
+T_{end}^n - T_{start}^n = D
+$$ 
+
+$D$ is hardcoded into the ProposerVM source code, and may only be changed by a required network upgrade.
+
+#### Changing the Epoch Duration
+
+Future network upgrades may change the value of $D$ to some new duration $D'$. $D'$ should not take effect until the end of the current epoch, rather than the activation time of the network upgrade that defines $D'$. This ensures an in progress epoch at the upgrade activation time cannot be less than both $D$ and $D'$.
+
+### Epoch Number
+
+Each epoch is associated with a monotonically increasing number, with $E_0$ being the epoch beginning at the activation time of the network upgrade that activates this ACP, and subsequent epochs incrementing the epoch number. Note that validators do not need to agree on epoch numbers, only the transition points between epochs, so no numbering scheme is provided in this ACP.
+
+### P-Chain Height
+
+As discussed above, the main [motivation](#motivation) for introducing epochs to the ProposerVM is to provide a fixed view of the P-Chain for the duration of the epoch. To achieve this, ProposerVM blocks are extended to include the current epoch's fixed P-Chain height, `PChainEpochHeight`, in addition to the latest P-Chain height, `PChainHeight`. The validator set at `PChainEpochHeight` should be used by the ProposerVM and the inner VM for the duration of the epoch. The first block of a new epoch will set the epoch's `PChainEpochHeight` to its parent's `PChainHeight`.
+
+This approach ensures that at any given height, the validator set to be used for the next block is known (this is a basic requirement for light clients). Put another way, within an epoch, the next block will use the current block's `PChainEpochHeight`, and at the boundary of the next epoch, the next block will use the current block's `PChainHeight`.
+
+## Backwards Compatibility
+
+This change requires a network upgrade and is therefore not backwards compatible.
+
+## Reference Implementation
+
+A reference implementation will be provided in [AvalancheGo](https://github.com/ava-labs/avalanchego), which must be merged before this ACP may be considered `Implementable`.
+
+## Security Considerations
+
+### Excessive Validator Churn
+
+The introduction of epochs concentrates validator set changes over the epoch's duration into a single block at the epoch's boundary. Excessive validator churn can cause consensus failures and other dangerous behavior, so it is imperative that the amount of validator weight change at the epoch boundary is limited. One strategy to accomplish this is to queue validator set changes and spread them out over multiple epochs. Another strategy is to batch updates to the same validator together such that increases and decreases to that validator's weight cancel each other out. Mechanisms to mitigate against this are outside the scope of this ACP and left to validator managers to implement. 
+
+## Open Questions
+
+What should the epoch duration $D$ be set to?
+
+Should the epoch numbering scheme be defined in this ACP?
+
+Should validator churn limits be implemented for the primary network to mitigate against [excessive validator churn](#excessive-validator-churn)?
+
+## Acknowledgements
+
+Thanks to [@iansuvak](https://github.com/iansuvak),  [@geoff-vball](https://github.com/geoff-vball), [@yacovm](https://github.com/yacovm), [@michaelkaplan13](https://github.com/michaelkaplan13), and [@StephenButtolph](https://github.com/StephenButtolph) for discussion and feedback on this ACP.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Epoch Duration
+
+Epochs have a constant duration $D$, such that for any given epoch $E_n$, the difference between its start and end times equals this duration: 
+
+$$
+T_{end}^n - T_{start}^n = D
+$$ 
+
+$D$ is hardcoded into the ProposerVM source code, and may only be changed by a required network upgrade.
```
geoff-vball, line 38: What if, instead of having a set epoch duration, where $T_{end}^n = T_{start}^{n+1}$, we used the timestamp of the block that sealed the previous epoch as the start time of the next epoch? This would eliminate `edge case 2`. In this case we should always be able to check the timestamp of the start of the epoch, because we already have to know `PChainEpochHeight`.

geoff-vball: I'm pretty sure it would eliminate the skew and actually simplify the current implementation. I'll write out an example...

D = 100

`Epoch 0`  `Tstart = 0` `Tend = 100` `PChainEpochHeight = 0`
Block 1 time = 0
Block 2 time = 57 
Block 3 time = 99
Block 4 time = 103 SEALS EPOCH 0

`Epoch 1` `Tstart = 103` `Tend = 203` `PChainEpochHeight = 4`
Block 5 time = 107
Block 6 time = 202
Block 7 time = 215 SEALS EPOCH 1

`Epoch 2` `Tstart = 215` `Tend = 315` `PChainEpochHeight = 7`
Block 8 time = 400 SEALS EPOCH 2

`Epoch 3` `Tstart = 400` `Tend = 500` `PChainEpochHeight = 8`

geoff-vball: All we need to check is if the timestamp of the parent is greater than `D` + `timestamp of the PChainEpochHeight` of the parent. If so, we advance the epoch.

cam-schultz: The edge case occurs if there's a long gap between the last block within the epoch bounds and the block that seals the epoch. Modifying your example:

Epoch 0 Tstart = 0 Tend = 100 PChainEpochHeight = 0
Block 1 time = 0
Block 2 time = 57
Block 3 time = 99
Block 4 time = **500** SEALS EPOCH 0

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,149 @@
...
+    if GetEpoch(parent.Timestamp()) != GetEpoch(grandParent.Timestamp()) {
+		// If the parent crossed the epoch boundary, then it sealed the previous epoch. The child
+		// is the first block of the new epoch, so should use the parent's P-Chain height.
+		return parent.PChainHeight()
+	}
+	// Otherwise, the parent did not seal the previous epoch, so the child should use the same
+	// epoch height. This is true even if the child crosses the epoch boundary, since sealing
+	// blocks are considered to be part of the epoch they seal.
+	return parent.PChainEpochHeight()
+}
```
michaelkaplan13, line 111:
```suggestion
type Block interface {
    Timestamp() time.Time
    PChainHeight() uint64
    PChainEpochHeight() uint64
}

// Returns the epoch number that the provided timestamp maps into.
func GetEpoch(timestamp time.Time) uint64 {
    ...
}

// [grandParent] is [parent]'s parent
func GetPChainEpochHeight(parent, grandParent Block) uint64 {
    if GetEpoch(parent.Timestamp()) != GetEpoch(grandParent.Timestamp()) {
        // If the parent crossed the epoch boundary, then it sealed the previous epoch. The child
        // is the first block of the new epoch, so should use the parent's P-Chain height.
        return parent.PChainHeight()
    }
    // Otherwise, the parent did not seal the previous epoch, so the child should use the same
    // epoch height. This is true even if the child crosses the epoch boundary, since sealing
    // blocks are considered to be part of the epoch they seal.
    return parent.PChainEpochHeight()
}
```

The tab spacing was off in the rendered view. I think sticking to spaces should fix it.

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,148 @@
...
+func GetPChainEpoch(parent, grandParent Block) (height, number uint64) {
+    if CrossedEpochBoundary(parent.Timestamp(), grandParent.Timestamp()) {
+		// If the parent crossed the epoch boundary, then it sealed the previous epoch. The child
+		// is the first block of the new epoch, so should use the parent's P-Chain height.
+		return parent.PChainHeight(), parent.EpochNumber() + 1
+	}
+	// Otherwise, the parent did not seal the previous epoch, so the child should use the same
+	// epoch height. This is true even if the child crosses the epoch boundary, since sealing
+	// blocks are considered to be part of the epoch they seal.
+	return parent.PChainEpochHeight(), parent.EpochNumber()
```
michaelkaplan13, line 107:
```suggestion
func CrossedEpochBoundary(timestamp1, timestamp2 time.Time) bool {}

// Calculates the P-Chain epoch height to be used for a given block, given
// its [parent] and [grandParent] blocks. [grandParent] is [parent]'s parent.
func GetPChainEpoch(parent, grandParent Block) (height, number uint64) {
    if CrossedEpochBoundary(parent.Timestamp(), grandParent.Timestamp()) {
        // If the parent crossed the epoch boundary, then it sealed the previous epoch. The child
        // is the first block of the new epoch, so should use the parent's P-Chain height.
        return parent.PChainHeight(), parent.EpochNumber() + 1
    }
    // Otherwise, the parent did not seal the previous epoch, so the child should use the same
    // epoch height. This is true even if the child crosses the epoch boundary, since sealing
    // blocks are considered to be part of the epoch they seal.
    return parent.PChainEpochHeight(), parent.EpochNumber()
```

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,148 @@
+| ACP | 181 |
+| :--- | :--- |
+| **Title** | P-Chain Epoched Views |
+| **Author(s)** | Cam Schultz [@cam-schultz](https://github.com/cam-schultz) |
+| **Status** | Proposed, Implementable, Activated, Stale ([Discussion](POPULATED BY MAINTAINER, DO NOT SET)) |
+| **Track** | Standards |
+
+## Abstract
+
+Proposes a standard P-Chain epoching scheme such that any VM that implements it has consistent view of the P-Chain for a known duration of time. This would enable VMs to optimize validator set retrievals, which currently must be done as often as every P-Chain block. This standard does *not* introduce epochs to the P-Chain's VM directly. Instead, it provides a standard that may be implemented by layers that inject P-Chain state into VMs, such as the ProposerVM.
```
michaelkaplan13, line 10:
```suggestion
Proposes a standard P-Chain epoching scheme such that any VM that implements it has a consistent view of the P-Chain for a known duration of time. This would enable VMs to optimize validator set retrievals, which currently must be done as often as every P-Chain block. This standard does *not* introduce epochs to the P-Chain's VM directly. Instead, it provides a standard that may be implemented by layers that inject P-Chain state into VMs, such as the ProposerVM.
```

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,148 @@
...
+E_n \coloneqq [ T_{start}^n,T_{end}^n )
+$$
+
+and the difference between $T_{end}^n$ and $T_{start}^n$ is $D$ for any $n$:
+
+$$
+T_{end}^n - T_{start}^n = D
+$$ 
+
+### Mapping Blocks to Epochs
```
michaelkaplan13, line 34: Would it make more sense to define epochs here by the blocks within them, rather than starting off with a definition based on time? I think we should avoid using terms like "start time" and "end time" here because the the actual start/end time of a given epoch is defined by the last block in the epoch before it and the final block included in it.

Thinking out loud, but we could do something like "An epoch $E_{N}$ is defined by the one or more blocks that have $N$ as their epoch number. A block's epoch number is defined according to the following:
- Epoch boundary time definition
- Block to epoch number mapping
- "

michaelkaplan13: I'm thinking more about this...

If an epoch N is defined by the blocks contained within it, should the "start time" of the epoch be the global time threshold that was crossed prior to the production of the block that sealed epoch N-1, or should it be the block timestamp of that block?

I'm imagine a rough recursive-like specification of something like:
- An epoch $E_N$ is defined by the one or more blocks that it contains.
- The first block contained in $E_N$ is the block immediately following the block that sealed epoch $E_{N-1}$. Call this sealing block $B_{S_{N-1}}$.
- The last block in $E_N$ is the block that seals $E_N$, called $B_{S_N}$, and is the defined as the first block with a timestamp of >= $T_{B_{S_{N-1}}} + D$

geoff-vball: [We discussed this a bit here](https://github.com/avalanche-foundation/ACPs/pull/181#discussion_r1974798153). I think I'm still generally in favour of the recursive definition.

cam-schultz: One nice property using the sealing block's timestamp as the epoch start time is that it guarantees the epoch duration will be at least $D$. That would improve the reliability of use cases such as ICM verification that require multiple network roundtrips to construct and submit the signed message.

The downside is that epoch transitions would not be coordinated across L1s, which in the current specification they would be. I don't know if this affects any of the targeted use cases, but wanted to call it out as a difference.

ACPs/181-proposervm-epochs/README.md
michaelkaplan13: Sorry I'm way late to this thread, but just had the same thought today as Geoff did back in Feb.

I don't think there are any edge cases with the approach Geoff proposed that don't also exist in the specification as it currently is, right? In the example you provided:
```
Epoch 0 Tstart = 0 Tend = 100 PChainEpochHeight = 0
Block 1 time = 0
Block 2 time = 57
Block 3 time = 99
Block 4 time = 500 SEALS EPOCH 0
```

Block 4 always must have the same `PChainEpochHeight` as block 1 either way, because there always needs to be a block that seals an epoch. The only difference is that the time duration `D` for epoch 2 would only begin at time = 500, rather than whatever the last fixed interval time going back to the network upgrade activation time is.

For example, say we `D=10`, and have:

```
Epoch 0 Tstart = 0
Block 1 time = 0
Block 2 time = 4
Block 3 time = 9
Block 4 time = 15
```

Block 4 seals epoch 0, but in the design as currently written, if the Block 5 comes in at time = 21, it would seal epoch 1 despite it's `PChainEpochHeight` only being 6 seconds old, since it is the "current" P-Chain height of Block 4. 

I think having the "end"/threshold time for epoch N being `(timestamp of block that sealed epoch N-1) + D` makes it more consistent such that the `PChainEpochHeight` is updated at most every `D` seconds since it last changed.

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,141 @@
...
+    return Epoch{
+        PChainHeight: parent.Epoch().PChainHeight
+        Number: parent.Epoch().Number
+        StartTime: parent.Epoch().StartTime
+    }
+}
+```
+
+- If the parent sealed its epoch, the current block [advances the epoch](#advancing-an-epoch), refreshing the epoch height, incrementing the epoch number, and setting the epoch starting time.
+- Otherwise, the current block uses the current epoch height, number, and starting time, regardless of whether it seals the epoch.
```
geoff-vball, line 107: I think it might be easier to reason about and implement this if we define the epoch starting time as the timestamp of the block that seals the previous epoch. Two reasons:

1. This makes the minimum number of blocks per epoch to be `1`. With the current approach, the minimum number of blocks per epoch is `2` because the first block sets the starting timestamp, which by definition cannot seal the epoch.
2. The parameters of the epoch would be completely defined by the block that seals the previous epoch - The P-Chain height, starting timestamp, and epoch number can all be fetched by looking at one block header. Otherwise we have to check two separate block headers to get all the epoch information.

cam-schultz: Only needing to reference the parent block is a nice property. There was some awkwardness around ordering in the reference implementation using the child's timestamp as the epoch start time. This would alleviate that.

The motivation to use the child's timestamp stemmed from having a predictable epoch length from a starting point denoted by a block acceptance, such that an epoch's duration could be reliably predicted by monitoring on-chain activity. However, the sealing block of the previous epoch gives the same property, with the only difference being that the epoch's duration depends on a block not in the epoch. I am not aware of any use cases for which this distinction is important.

In short, I agree with your comments and will make the change. Thanks for the suggestion.

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,141 @@
...
+### Epoch Definition
+
+An epoch is defined as a contiguous range of blocks that share the same three values:
+- An Epoch Number
+- An Epoch P-Chain Height
+- An Epoch Start Time
+
+Let $E_N$ denote an epoch with epoch number $N$. $E_N$'s start time is denoted as $T_{start}^N$, and its P-Chain height as $P_N$. 
+
+$E_0$ is defined as the epoch whose start time $T_{start}^N$ is the activation time of the network upgrade that activates this ACP.
```
michaelkaplan13, line 33:
```suggestion
$E_0$ is defined as the epoch whose start time $T_{start}^0$ is the block timestamp of the block that activates this ACP (i.e. the first block at or following this ACP's activation timestamp).
```

ACPs/181-proposervm-epochs/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Fixing the P-Chain Height
+
+When building a block, Avalanche blockchains use the P-Chain height [embedded in the block](#assumptions) to determine the validator set. If instead the epoch P-Chain height is used, then we can ensure that when a block is built, the validator set to be used for the next block is known. To see this, suppose block $b_m$ seals epoch $E_N$. Then the next block, $b_{m+1}$ will begin a new epoch, $E_{N+1}$ with $P_{N+1}$ equal to $b_m$'s P-Chain height, $p_m$. If instead $b_m$ does not seal $E_N$, then $b_{m+1}$ will continue to use $P_{N}$. Both candidates for $b_{m+1}$'s P-Chain height ($p_m$ and $P_N$) are known at $b_m$ build time.
+
+## Backwards Compatibility
+
+This change requires a network upgrade and is therefore not backwards compatible.
```
michaelkaplan13, line 60: I'd be great to add a second paragraph here calling out any known incompatibilities this upgrade will cause. The only one I can think of right now is that relayers of ICM message will need to account for the destination chain's epoching scheme in order to be able to construct an aggregate signature that will be verified properly

ACPs/181-proposervm-epochs/README.md
michaelkaplan13: Bumping this suggestion.

michaelkaplan13 (approved):
Two small suggestions but otherwise LGTM

PR #182, opened 2025-02-18
Updated K to closer match current rate of change.

alarso16:
With $$K=43*T$$, the fee can double about every 30 seconds. However, this varies greatly from the C-Chain and Ethereum's current rate, which can double around every 50 seconds and 70 seconds, respectively. To keep the reactivity to demand changes relatively similar to the current mechanism, we should choose a $$K$$ that is close to these values (i.e. 1 minute).


StephenButtolph (approved):
@alarso16 could you add yourself to the `Acknowledgments` section?

Other than that lgtm

PR #186, opened 2025-03-12
[ACP-185] Update Validator Manager

martineckardt:
Proposes a new P-Chain tx type to update the validator manager of a Subnet


ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+| :------------ | :-------------------------------------------------- |
+| **Title**     | Update Validator Manager                            |
+| **Author(s)** | [Martin Eckardt](https://github.com/martineckardt)  |
+| **Status**    | Proposed                                            |
+| **Track**     | Standards                                           |
+| **Related**   | [ACP-77](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/77-reinventing-subnets/README.md)       |
+
+## Abstract
+
+This ACP proposes an extension to ACP-77 by introducing a new transaction type `UpdateValidatorManagerTx` to facilitate validator manager migration and upgrading without requiring proxy contracts or network upgrades. This enhancement allows L1s to:
```
meaghanfitzgerald, line 11:
```suggestion
This ACP proposes an extension to [ACP-77](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/77-reinventing-subnets/README.md)  by introducing a new transaction type `UpdateValidatorManagerTx` to facilitate validator manager migration and upgrading without requiring proxy contracts or network upgrades. This enhancement allows L1s to:
```

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+
+- Migrate validator management between different chains (e.g., from C-Chain to a native L1)
+- Recover from misconfigurations where validator managers were set to inaccessible chains
```
meaghanfitzgerald, line 14: is this referring to an instance where the validator management contract is deployed on a chain that may have stopped building blocks? if so I suggest:
```suggestion
- Recover from incidents where the validator manager address was set to a chain that has halted or become corrupted
```

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+- Avoid the need for proxy contracts and their associated security concerns
+- Implement smooth upgrade paths without requiring network upgrades
+
+By providing a flexible mechanism for validator manager migration, this proposal significantly enhances the sovereignty and adaptability of Avalanche L1s.
```
meaghanfitzgerald, line 18:
```suggestion
By providing a flexible mechanism for validator manager migration, this proposal significantly enhances the sovereignty, security, and adaptability of Avalanche L1s.
```

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Specification
+
+To address these challenges, this ACP proposes a new transaction type `UpdateValidatorManagerTx` that allows an L1 to update its validator manager without requiring proxy contracts or network upgrades.
+
+### New P-Chain Transaction Type
+
+#### `UpdateValidatorManagerTx`
+
+The `UpdateValidatorManagerTx` will allow an L1 to update its validator manager contract address and chain ID:
```
meaghanfitzgerald, line 61:
```suggestion
The `UpdateValidatorManagerTx` will allow an L1 to update its validator manager contract address and `ChainID`:
```

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+    // Metadata, inputs and outputs
+    BaseTx
+    // ID of the L1 to update the validator manager for
+    // Restrictions:
+    // - Must not be the Primary Network ID
+    // - Must be an L1 (previously converted via ConvertSubnetToL1Tx)
+    SubnetID ids.ID `json:"subnetID"`
+    // New BlockchainID where the validator manager will live
+    // Restrictions:
+    // - Cannot be the P-Chain chain ID
```
meaghanfitzgerald, line 74:
```suggestion
    // - Cannot be the P-Chain ChainID
```

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+    // SubnetID identifies the L1 whose validator manager was updated
+    SubnetID ids.ID `json:"subnetID"`
+    // OldChainID identifies the blockchain where the old validator manager lived
+    OldChainID ids.ID `json:"oldChainID"`
+    // OldAddress is the address of the old validator manager
+    OldAddress []byte `json:"oldAddress"`
+    // NewChainID identifies the blockchain where the new validator manager lives
+    NewChainID ids.ID `json:"newChainID"`
+    // NewAddress is the address of the new validator manager
+    NewAddress []byte `json:"newAddress"`
```
meaghanfitzgerald, line 139: Not sure that both the old and the new chain information needs to be conveyed by the update message, it should be enough to provide a boolean and the current active manager and address. This follows the precedent set in [ACP-77](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#l1validatorregistrationmessage) by `L1ValidatorRegistrationMessage`, wherein, "`registered` is a boolean representing the status of the validationID. If true, the validationID corresponds to a validator in the current validator set. If false, the validationID does not correspond to a validator in the current validator set, and never will in the future."

```suggestion
    // AddressUpdateID identifies attempted change in validator manager address and/or chainID for the message
    AddressUpdateID ids.ID `json:"addressUpdateID"`
    // Updated returns true if the validator manager `Address` has been updated
    Updated boolean `json:"updated"`
    // ChainID identifies the blockchainID of the active validator manager contract
    ChainID ids.ID `json:"chainID"`
    // Address identifies the contract address of the active validator manager contract
    Address []byte `json:"address"`
```
In this instance, `Updated` is a boolean representing the status of the `AddressUpdateID`. If true, the addressUpdateID corresponds to the current active `Address` and `ChainID`. If false, the `AddressUpdateID` does not correspond to these values, and never will in the future.

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+    // ValidatorList is the serialized list of validators at the time of update
+    ValidatorList []ValidatorData `json:"validatorList"`
```
meaghanfitzgerald, line 141: shouldn't this be
```
Validators []L1Validator `json:"validators"`
```

meaghanfitzgerald:
One large question I have here is, how will the state of the validators and their rewards/ validation status be preserved when the address, and ultimately the storage slots of all the validator reward-relevant information is changed by action on a separate chain? How will we avoid race conditions between the changing state of the old validator manager contract and the valid `Address` of the VMC according to the P-Chain i.e. in the instance where a validator may be in the process of registering with the old VMC while the address is being changed on the P-Chain?

ACPs/186-update-validator-manager/Readme.md
martineckardt: I was also referring to situations where the validator manager chain ID was set to the P-Chain. Another problem would arise if the validator manager was pointing to the wrong address on the C-Chain . If the latter happened on your L1 you could put a validator manager in the state there with a network upgrade, but that cannot be done on the C-Chain.

ACPs/186-update-validator-manager/Readme.md
martineckardt: I was thinking to include both to mage the migration from one validator manager contract to the other easier. If this was a PoS staking manager that was holding funds, having this very explicit message would allow designing a trustless migration where in a single transaction the old validator manager can be deactivated and instructed to move the funds to the validator manager. If there is ambiguity of what exactly happened on the P-Chain the migration process may need a trusted actor

ACPs/186-update-validator-manager/Readme.md
martineckardt: It reuses the same data type as the conversion message. Updated the ACP to use the exact naming and refer to the current implementation

https://github.com/ava-labs/avalanchego/blob/d36f7d392bdae99a68ced5c4f6ef345ff3ce25da/vms/platformvm/warp/message/subnet_to_l1_conversion.go#L25C19-L25C52

martineckardt:
> One large question I have here is, how will the state of the validators and their rewards/ validation status be preserved when the address, and ultimately the storage slots of all the validator reward-relevant information is changed by action on a separate chain? How will we avoid race conditions between the changing state of the old validator manager contract and the valid `Address` of the VMC according to the P-Chain i.e. in the instance where a validator may be in the process of registering with the old VMC while the address is being changed on the P-Chain?

I think the PoA case is pretty straight forward. It get's difficult once there is no owner and there are tokens involved. 

Ideally, the PoS VMC is involved in that flow. High level I am thinking about an `initiateValidatorManagerUpdate` method that locks the current state for a certain time for example. A requirement could be that there are no pending registrations, delegation and removals at that time. We could also possibly add the state root of the locked VMC state in the `UpdateValidatorManagerMessage` so it can be reconstructed on the new validator manager.

We also need to account for a scenario where the VMC is not available anymore, so we can't make the step above required. The L1 validator set should always have the power.

ACPs/186-update-validator-manager/Readme.md
```diff
@@ -0,0 +1,219 @@
...
+
+1. **Centralization of Power**: The proxy owner has extensive control over the validator set through the ability to change the validator manager implementation - essentially equivalent to a Proof-of-Authority admin even for Proof-of-Stake chains.
+
+2. **Chain Migration Barriers**: The validator manager cannot move between chains. Projects like Beam that wish to start by managing their validator set on the C-Chain before eventually moving to their own L1 face a significant roadblock.
+
+3. **No Recovery Path for Misconfiguration**: There is currently no recovery path for L1s that accidentally set their validator manager chainID to a chain they cannot control. While L1s can modify contract code on their own chain via a network upgrade, there is no recovery mechanism when the manager is set to an external chain. There have been instances of production L1s that nearly set their validator manager chainID to the P-Chain chainID, which would have rendered them permanently unrecoverable.
```
geoff-vball, line 49: Nearly set the chainID to the P-Chain chainID in the Validator Manager contract, or in the tx to the P-Chain? In either case, should we add a defensive check against this since it's invalid?

geoff-vball (commented):
What happens to in-flight weight updates? If a delegator stakes to a validator, and then the validator manager contract is migrated before the weight update message can be delivered, will that weight change still be able to be applied?

ACPs/186-update-validator-manager/Readme.md
martineckardt: This specific case was on the P-Chain. I am chatting with Stephen about a separate ACP to protect against this

PR #189, opened 2025-03-13
[ACP-189] Seamless L1 Creation

martineckardt:

PR #191, opened 2025-03-24
[ACP-191] Seamless L1 Creation

martineckardt:
Proposing a new P-Chain tx type CreateL1Tx to create an L1 in one transaction instead of three.

Replacing [ACP-189](https://github.com/avalanche-foundation/ACPs/pull/189)

### DIFF
--- a/ACPs/191-seamless-l1-creation/README.md
+++ b/ACPs/191-seamless-l1-creation/README.md
@@ -1,0 +1,130 @@
+| ACP | 189 |
+| :- | :- |
+| **Title** | Seamless L1 Creations (CreateL1Tx) |
+| **Author(s)** | Martin Eckardt ([@martineckardt](https://github.com/martineckardt)), Aaron Buchwald ([@aronbuchwald](https://github.com/aaronbuchwald)), Michael Kaplan ([@michaelkaplan13 ](https://github.com/michaelkaplan13)), Meag FitzGerald ([@meaghanfitzgerald](https://github.com/meaghanfitzgerald)) |
+| **Status** | Proposed |
+| **Track** | Standards |
+
+## Abstract
+
+This ACP introduces a new P-Chain transaction type called `CreateL1Tx` that simplifies the creation of Avalanche L1s. It consolidates three existing transaction types (`CreateSubnetTx`, `CreateChainTx`, and `ConvertSubnetToL1Tx`) into a single atomic operation. This streamlines the L1 creation process, removes the need for the intermediary Subnet creation step, and eliminates the management of temporary SubnetAuth credentials.
+
+## Motivation
+
+[ACP-77](../77-reinventing-subnets/README.md) introduced Avalanche L1s, providing greater sovereignty and flexibility compared to Subnets. However, creating an L1 currently requires a three-step process:
+
+1. `CreateSubnetTx`: Create the Subnet record on the P-Chain and specify the `SubnetAuth `
+2. `CreateChainTx`: Add a blockchain to the Subnet (can be called multiple times)  
+3. `ConvertSubnetToL1Tx`: Convert the Subnet to an L1, specifying the initial validator set and the validator manager location
+
+This process has several drawbacks:
+
+* It requires orchestrating three separate transactions that could be handled in one.
+* The `SubnetAuth` must be managed during creation but becomes irrelevant after conversion.
+* The multi-step process increases complexity and potential for errors.
+* It introduces unnecessary state transitions and storage overhead on the P-Chain.
+
+By introducing a single `CreateL1Tx` transaction, we can simplify the process, reduce overhead, and improve the developer experience for creating L1s.
+
+## **Specification**
+
+### New Transaction Type
+
+The following new transaction type is introduced:
+
+```go
+// ChainConfig represents the configuration for a chain to be created
+type ChainConfig struct {
+    // A human readable name for the chain; need not be unique
+    ChainName string `serialize:"true" json:"chainName"`
+    // ID of the VM running on the chain
+    VMID ids.ID `serialize:"true" json:"vmID"`
+    // IDs of the feature extensions running on the chain
+    FxIDs []ids.ID `serialize:"true" json:"fxIDs"`
+    // Byte representation of genesis state of the chain
+    GenesisData []byte `serialize:"true" json:"genesisData"`
+}
+
+// CreateL1Tx is an unsigned transaction to create a new L1 with one or more chains
+type CreateL1Tx struct {
+    // Metadata, inputs and outputs
+    BaseTx `serialize:"true"`
+    
+    // Chain configurations for the L1 (can be multiple)
+    Chains []ChainConfig `serialize:"true" json:"chains"`
+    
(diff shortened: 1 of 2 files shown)


ACPs/191-seamless-l1-creation/README.md
```diff
@@ -0,0 +1,130 @@
...
+1. Create a new L1.
+2. Create chain records for each chain configuration in the `Chains` array.  
+3. Set up the L1 validator manager with the specified `ManagerChainID` and `ManagerAddress`.  
+4. Register the initial validators specified in the `Validators` array.
+
+### IDs
+
+* `subnetID`: The `subnetID` of the L1 is the transaction hash.
+* `blockchainID`: the `blockchainID` for each blockchain is is defined as the SHA256 hash of the 37 bytes resulting from concatenating the 32 byte subnetID with the `0x00` byte and the 4 byte chainIndex (index in the `Chains` array within the transaction)
+* `validationID`: The `validationID` for the initial validators added through CreateL1Tx is defined as the SHA256 hash of the 37 bytes resulting from concatenating the 32 byte subnetID with the `0x01` byte and 4 byte validatorIndex (index in the `Validators` array within the transaction).
```
meaghanfitzgerald, line 82: > Summarizing and continuing this conversation from the [preceding PR](https://github.com/avalanche-foundation/ACPs/pull/189#discussion_r2001534277)

0x00 and 0x01 are used here to distinguish between blockchain IDs and validation IDs.

@michaelkaplan13: 

> Ideally, I think we should keep the validation ID definition so that the new L1 creation transaction is compatible with the existing validator manager contracts...I'll try to think of cleanest way to make that possible.


@meaghanfitzgerald: 

> Could blockchainID = sha256(subnetID + chainIndex), and the validationID = sha256(subnetID + (validatorIndex + (length(Chains)-1)))?
> 
> alternatively, if the validator manager contracts depend on assumptions/knowledge of the exact index of a validator in the Validators array, we could flip the above order, i.e:
> validationID = sha256(subnetID + validatorIndex), and the blockchainID = sha256(subnetID + (chainIndex + (length(Validators)-1)))?
> _Or_
> This transaction could enforce that only single-chain L1 networks are created.

@martineckardt:

> The only way of getting rid of the byte to distinguish would be using the same counter/index as Meag suggested for both initial validators and blockchains. The trade-off is between having less complexity to generate the chain IDs (you don't need to know how many validators are included in the convert tx, you can just count from 0) vs. the introduction of this additional byte.
> Don't think we should enforce single-chain L1 networks just here. If this is something we generally think is desirable, we should also enforce this is in the CreateChainTx. I would want to avoid a situation where we have to document that the CreateL1Tx works for most cases, but you need to do the legacy flow in some cases (multiple chains).

ACPs/191-seamless-l1-creation/README.md
```diff
@@ -0,0 +1,130 @@
+| ACP | 189 |
```
meaghanfitzgerald, line 1:
```suggestion
| ACP | 191 |
```
to match the PR number?

ACPs/191-seamless-l1-creation/README.md
```diff
@@ -0,0 +1,130 @@
...
+This process has several drawbacks:
+
+* It requires orchestrating three separate transactions that could be handled in one.
+* The `SubnetAuth` must be managed during creation but becomes irrelevant after conversion.
+* The multi-step process increases complexity and potential for errors.
+* It introduces unnecessary state transitions and storage overhead on the P-Chain.
+
+By introducing a single `CreateL1Tx` transaction, we can simplify the process, reduce overhead, and improve the developer experience for creating L1s.
+
+## **Specification**
```
michaelkaplan13, line 29:
```suggestion
## Specification
```

ACPs/191-seamless-l1-creation/README.md
```diff
@@ -0,0 +1,130 @@
+| ACP | 191 |
+| :- | :- |
+| **Title** | Seamless L1 Creations (CreateL1Tx) |
+| **Author(s)** | Martin Eckardt ([@martineckardt](https://github.com/martineckardt)), Aaron Buchwald ([@aaronbuchwald](https://github.com/aaronbuchwald)), Michael Kaplan ([@michaelkaplan13](https://github.com/michaelkaplan13)), Meaghan FitzGerald ([@meaghanfitzgerald](https://github.com/meaghanfitzgerald)) |
+| **Status** | Proposed |
+| **Track** | Standards |
+
+## Abstract
+
+This ACP introduces a new P-Chain transaction type called `CreateL1Tx` that simplifies the creation of Avalanche L1s. It consolidates three existing transaction types (`CreateSubnetTx`, `CreateChainTx`, and `ConvertSubnetToL1Tx`) into a single atomic operation. This streamlines the L1 creation process, removes the need for the intermediary Subnet creation step, and eliminates the management of temporary SubnetAuth credentials.
```
michaelkaplan13, line 10:
```suggestion
This ACP introduces a new P-Chain transaction type called `CreateL1Tx` that simplifies the creation of Avalanche L1s. It consolidates three existing transaction types (`CreateSubnetTx`, `CreateChainTx`, and `ConvertSubnetToL1Tx`) into a single atomic operation. This streamlines the L1 creation process, removes the need for the intermediary Subnet creation step, and eliminates the management of temporary `SubnetAuth` credentials.
```

ACPs/191-seamless-l1-creation/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Motivation
+
+[ACP-77](../77-reinventing-subnets/README.md) introduced Avalanche L1s, providing greater sovereignty and flexibility compared to Subnets. However, creating an L1 currently requires a three-step process:
+
+1. `CreateSubnetTx`: Create the Subnet record on the P-Chain and specify the `SubnetAuth `
```
michaelkaplan13, line 16:
```suggestion
1. `CreateSubnetTx`: Create the Subnet record on the P-Chain and specify the `SubnetAuth`
```

ACPs/191-seamless-l1-creation/README.md
michaelkaplan13: I think there are probably a few options:
1. Let the blockchain IDs and validations overlap. They're maintained in different places in the P-Chain state, so there shouldn't be any technical issue with them being the same. It's just potentially odd/confusing UX.
2. Keep `validationID`s as `SHA256(subnetID,validatorIndex)`, and redefine `blockchainID` to be `SHA256(subnetID,0x00,blockchainIndex`. This should be sufficient for making them unique without modifying the existing `validationID` calculation, and the `blockchainID` definition is changing either way for this flow.
3. Go with what is currently proposed here, and update the validation manager contracts to account for it. This should be technically complex, but people would have to be aware of it in order to chose the correct version of the contracts based on how they created their L1 on the P-Chain, which seems undesirable.

meaghanfitzgerald:
Here is the sdk one could use to manually generate node credentials: https://github.com/ava-labs/avalanche-cli/blob/main/cmd/nodecmd/create.go#L977

michaelkaplan13 (approved):

PR #205, opened 2025-06-03
Document ACP-176 block building heuristic

StephenButtolph:
Avalanchego has adopted this block building heuristic (bounding the delay up to 15M gas to align with pre-ACP-176 expectations) here: https://github.com/ava-labs/coreth/pull/995


ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md
```diff
@@ -133,6 +133,9 @@ While Avalanche Network Clients can set default configuration values for the des
...
 
+> [!NOTE]
+> After the activation of this ACP, concerns were raised around the latency of inclusion for large transactions when the fee is increasing. To address these concerns, block producers SHOULD only produce blocks when there is sufficient capacity to include large transactions. Prior to this ACP, the maximum size of a transaction was $15$ million gas. Therefore, the recommended heuristic is to only produce blocks when there is at least $\min(8 \cdot T, 15 \text{ million})$ capacity. _At the time of writing, this ensures transactions with up to $12.8$ million gas will be able to bid for block space._
```
meaghanfitzgerald, line 137: image

```suggestion
> After the activation of this ACP, concerns were raised around the latency of inclusion for large transactions when the fee is increasing. To address these concerns, block producers SHOULD only produce blocks when there is sufficient capacity to include large transactions. Prior to this ACP, the maximum size of a transaction was $15$ million gas. Therefore, the recommended heuristic is to only produce blocks when there is at least $\min(8 \cdot T, 15 \text{ million})$ capacity. _At the time of writing, this ensures transactions with up to 12.8 million gas will be able to bid for block space._
```

StephenButtolph: GitHub's $\LaTeX$ sucks

PR #214, opened 2025-06-19
[ACP-194] update named-block label definitions

ARR4N:
There has been some confusion about whether ACP-194 (a) modifies finality guarantees, and (b) forces users to wait until settlement (i.e. the _safe_ block). Both of these are incorrect interpretations.

Having APIs support both the _safe_ and _finalized_ block labels, with the same meaning, contributed significantly to (a) because of the word _finalized_. As they were redundant, use of _finalized_ is removed.

Concerns around (b) are clarified with an explanation of the meaning of _safe_ (absolute and with respect to hard-drive corruption), with guidance on its usage and alternatives to achieve faster guarantees.

### DIFF
--- a/ACPs/194-streaming-asynchronous-execution/README.md
+++ b/ACPs/194-streaming-asynchronous-execution/README.md
@@ -403,15 +403,21 @@ In particular, the API method `eth_getBlockReceipts` MUST return the receipts co
 #### Named blocks
 
 The Ethereum Mainnet APIs allow for retrieving blocks by named parameters that the API server resolves based on their consensus mechanism.
-Other than the _earliest_ (genesis) named block, which MUST be interpreted in the same manner, all other named blocks are mapped to SAE in terms of the _execution_ status of blocks and MUST be interpreted as follows:
+C-Chain does not currently support _pending_, _safe_, and _finalized_ labels.
+This ACP modifies the meaning of _latest_ and adopts an altered definition of _safe_, which MUST each be resolved as:
 
- * _pending_: the most recently _accepted_ block;
- * _latest_: the block that was most recently _executed_;
- * _safe_ and _finalized_: the block that was most recently _settled_.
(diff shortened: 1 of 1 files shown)


ACPs/194-streaming-asynchronous-execution/README.md
```diff
@@ -403,15 +403,21 @@ In particular, the API method `eth_getBlockReceipts` MUST return the receipts co
 #### Named blocks
 
 The Ethereum Mainnet APIs allow for retrieving blocks by named parameters that the API server resolves based on their consensus mechanism.
-Other than the _earliest_ (genesis) named block, which MUST be interpreted in the same manner, all other named blocks are mapped to SAE in terms of the _execution_ status of blocks and MUST be interpreted as follows:
+C-Chain does not currently support _pending_, _safe_, and _finalized_ labels.
```
StephenButtolph, line 406: I don't think this comment is correct. The C-chain aliases `pending`, `latest`, `safe`, and `finalized` as the last accepted block.

StephenButtolph:
We had discussed this offline, but just to write it down: There is a pretty high probability that we'll need to support all of the labels for compatibility with standard ethereum tooling.

ARR4N:
> We had discussed this offline, but just to write it down: There is a pretty high probability that we'll need to support all of the labels for compatibility with standard ethereum tooling.

What do we want to do with "finalized" then? If we make it the same as "latest" then the ordering is different to what people expect. If it's an alias of "safe" (the natural definition in my opinion) then we have the same problem of people incorrectly thinking that "now Avalanche has 5s finality".

ACPs/194-streaming-asynchronous-execution/README.md
ARR4N: Fixed.

StephenButtolph:
I think we will probably need to maintain the same ordering that tooling expects and just deal with any misconseptions that arise from it

PR #215, opened 2025-06-24
[ACP-194] SAE: Simplify Delta t assignment

martineckardt:
Replaced the original expression `Δt := max(te, tb) − te` with the simplified and equivalent form `Δt := max(0, tb − te)`. This improves clarity by directly expressing the non-negative delay between tb and te, and avoids unnecessary use of `max(te, tb)`.


ARR4N (approved):
Thank you! Please add yourself to the acknowledgments section.

PR #222, opened 2025-07-09
ACP-204: Update Precompile Address

michaelkaplan13:
Rather than use the Avalanche precompile address space (i.e. 0x02...), I think it would be better to put this precompile at the same address as it already is available at in other ecosystems.


michaelkaplan13:
@ARR4N / @scammi I can't add you as reviewers directly, but thoughts on aligning the precompile address with here with the one specified by the RIP? 🙏

ARR4N (approved):

PR #224, opened 2025-07-11
[ACP-224] Add Dynamic Gas Limit and Price Discovery to Subnet-EVM

ceyonur:
This PR adds a proposal to enable acp-176 style fee mechanism and dynamic gas limits to Subnet-EVM along with a new fee manager precompile.

### DIFF
--- a/ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
+++ b/ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
@@ -68,7 +68,7 @@ Prior to this ACP, the Subnet-EVM fee configuration and fee manager precompile u
 
 ACP-176 will make `GasLimit` and `BaseFeeChangeDenominator` configurations obsolete in Subnet-EVM.
 
-`TargetBlockRate`, `MinBlockGasCost`, `MaxBlockGasCost`, and `BlockGasCostStep` will be also removed by [ACP-226](https://github.com/michaelkaplan13/ACPs/blob/main/ACPs/226-dynamic-minimum-block-times/README.md).
+`TargetBlockRate`, `MinBlockGasCost`, `MaxBlockGasCost`, and `BlockGasCostStep` will be also removed by [ACP-226](https://github.com/avalanche-foundation/ACPs/tree/ce51dfab/ACPs/226-dynamic-minimum-block-times).
 
 `MinGasPrice` is equivalent to `M` in ACP-176 and will be used to set the minimum gas price for ACP-176. This is similar to `MinBaseFee` in old Subnet-EVM fee configuration, and roughly gives the same effect. Currently the default value is `25 * 10^-18` (25 nAVAX/Gwei). This default will be changed to the minimum possible denomination of the native EVM asset (1 Wei), which is aligned with the C-Chain.
 


ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+ACP-194 (SAE) depends on having a gas target and capacity mechanism aligned with ACP-176 to maintain consistency across Subnet-EVM and Coreth. It would be very difficult to use the existing windower fee mechanism in conjunction with SAE, and this would also require maintaining multiple mechanisms on a go-forwards basis. As such, adding ACP-176 into subnet-evm is (more or less) a requirement for L1s to be able to use SAE in the future.
+
+While both ACP-176 and ACP-194 will be required upgrades for L1s, this ACP aims to still provide similar controls for chains with a new precompile. To do so, we will add a new dynamic fee configuration and fee manager precompile that maps well into the ACP-176 mechanism, while still allowing admins to adjust the fee parameters dynamically and unilaterally.
+
+## Specification
+
+### C-Chain Parameters (ACP-176)
+
+This ACP aims to use same parameters as in [ACP-176 configuration parameters](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md#configuration-parameters):
```
michaelkaplan13, line 25:
```suggestion
### ACP-176 Parameters

This ACP uses same parameters as in the [ACP-176 specification](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md#configuration-parameters), and allows their values to be configured on a chain-by-chain basis. The parameter list and current values used by the C-Chain are as follows:
```

Wording nit: It sounded like the C-Chain configuration would be used across the board for all chains, which I realize is not the case. I think this clarifies it a bit.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+| C | maximum gas capacity | 10*T |
+| P | minimum target gas consumption per second | 1,000,000 |
+| D | target gas consumption rate update constant | 2^25 |
+| Q | target gas consumption rate update factor change limit | 2^15 |
+| M | minimum gas price | 1∗10^−18 AVAX |
+| K | initial gas price update factor | 87*T |
+
+### Prior Subnet-EVM Fee Configuration Parameters
+
+In the prior (before this ACP) Subnet-EVM fee configuration and in the fee manager precompile, the following parameters are used to control the fee mechanism:
```
michaelkaplan13, line 40:
```suggestion
Prior to this ACP, the Subnet-EVM fee configuration and fee manager precompile used the following parameters to control the fee mechanism:
```

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+  Sets the max amount of gas consumed per block.
+
+**TargetBlockRate**:
+  Sets the target rate of block production in seconds used for fee adjustments. If the actual block rate is faster than this target, block gas cost will be increased, and vice versa.
+
+**MinBaseFee**:
+  The minimum base fee sets a lower bound on the EIP-1559 base fee of a block. Since the block's base fee sets the minimum gas price for any transaction included in that block, this effectively sets a minimum gas price for any transaction.
+
+**TargetGas**:
+  Specifies the targeted amount of gas (including block gas cost) to consume within a rolling 10s window. When the dynamic fee algorithm observes that network activity is above/below the TargetGas, it increases/decreases the base fee proportionally to how far above/below the target actual network activity is.
```
michaelkaplan13, line 52:
```suggestion
  Specifies the targeted amount of gas (including block gas cost) to consume within a rolling 10s window. When the dynamic fee algorithm observes that network activity is above/below the `TargetGas`, it increases/decreases the base fee proportionally to how far above/below the target actual network activity is.
```

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+**MinBlockGasCost**:
+  Sets the minimum amount of gas to charge for the production of a block.
+
+**MaxBlockGasCost**:
+  Sets the maximum amount of gas to charge for the production of a block.
+
+**BlockGasCostStep**:
+  Determines how much to increase/decrease the block gas cost depending on the amount of time elapsed since the previous block. If the block is produced at the target rate, the block gas cost will stay the same as the block gas cost for the parent block. If it is produced faster/slower, the block gas cost will be increased/decreased by the step value for each second faster/slower than the target block rate accordingly.
+  Note: if the BlockGasCostStep is set to a very large number, it effectively requires block production to go no faster than the TargetBlockRate.
```
michaelkaplan13, line 65:
```suggestion
  Note: if the `BlockGasCostStep` is set to a very large number, it effectively requires block production to go no faster than the `TargetBlockRate`.
```

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+  Ex: if a block is produced two seconds faster than the target block rate, the block gas cost will increase by 2 * BlockGasCostStep.
```
michaelkaplan13, line 66:
```suggestion
  Ex: if a block is produced two seconds faster than the target block rate, the block gas cost will increase by `2 * BlockGasCostStep`.
```

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+    returns (
+      uint256 targetGas;
+      uint256 minGasPrice;
+      uint256 maxCapacityFactor;
+      uint256 timeToDouble;
+      uint256 targetBlockRate;
+      uint256 minBlockGasCost;
+      uint256 maxBlockGasCost;
+      uint256 blockGasCostStep;
+    );
```
michaelkaplan13, line 172:
```suggestion
  // Set fee config fields to contract storage
  function setFeeConfig(
    ACP176FeeConfig calldata config
  ) external;

  // Get fee config from the contract storage
  function getFeeConfig()
    external
    view
    returns (
      ACP176FeeConfig memory config
    );
```

Is it possible to just pass the config struct to the precompile interface? If so, I think that's the cleaner approach rather than having 8 individual params

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+```mermaid
+flowchart TD
+    A[ACP-224 activated] --> B{Is ACP176FeeManager precompile active?}
+
+    B -- Yes --> C[Use targetGas from precompile storage]
+
+    B -- No --> D{Is gas-target set in node config?}
+    D -- Yes --> E[Use gas-target from node config]
+
+    D -- No --> F{Is targetGas set in chainConfig.acp176FeeConfig?}
```
michaelkaplan13, line 190: What's the "node config" value here? If you mean the value set for the node for the C-Chain, I don't think we should be re-using it here, and could probably simplify this by just looking at the chain config.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+    F -- Yes --> G[Use targetGas from chain config]
+
+    F -- No --> H[Use default targetGas e.g. 2 million]
```
michaelkaplan13, line 193: For ACP-176, the implementation actually uses the current on-chain value as the "default" effectively abstaining from having a preference if none is set. I think we should do the same here.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+      uint256 maxBlockGasCost;
+      uint256 blockGasCostStep;
+    );
+
+  // Get the last block number changed the fee config from the contract storage
+  function getFeeConfigLastChangedAt() external view returns (uint256 blockNumber);
+}
+```
+
+The precompile will have a `targetGas` parameter that is used to set the target gas consumed per second for ACP-176. This will override the `gas-target` in node config. If the new fee manager precompile is not activated, validators can keep using `gas-target` in node config to set the target gas consumed per second (as in C-Chain).
```
michaelkaplan13, line 179: Is following of the `targetGas` (or more technically `q` under the hood) set in the `ACP176FeeManager` precompile something that is enforced by block verification? 

For instance, primary network validators today can set their preferred target gas to any value they want, but the effect target gas value for the chain can only be altered by a fixed factor in each subsequent block, and the new effect target gas value being within the allowed bounds is verified in block verification.

For chains with `ACP176FeeManager` precompile activated, can we somehow just enforce that the target gas value is always equal to the latest accept value provided to the precompile? So if `setFeeConfig` is called in block N, block N+1 must have the effective `targetGas` specified in the final call in block N.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+    D -- No --> F{Is targetGas set in chainConfig.acp176FeeConfig?}
+    F -- Yes --> G[Use targetGas from chain config]
+
+    F -- No --> H[Use default targetGas e.g. 2 million]
+```
+
+## Backwards Compatibility
+
+ACP-224 will require a network update in order to activate the new fee mechanism. Another activation will also be required to activate the new fee manager precompile. The activation of precompile should never occur before the activation of ACP-224 (the fee mechanism) since the precompile depends on ACP-224’s fee update logic to function correctly.
```
michaelkaplan13, line 198: Could these activations be scheduled one immediately (or one second) after each other? Or is there an edge case if there is no blocks produced in between their activations?

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+* Does it make sense dynamically changing the D and Q parameters? Currently they give each block builder the ability to adjust the value of $T$ on the rate of change of by roughly $\frac{1}{1024}$ of its current value.
+
+* Similarly does it make sense to dynamically change the `RMult` parameter? Currently it is set to 2, making `R = 2 * T` which means gas price can increase and decrease at the same rate.
+
+* Should we rather name the precompile as `ACP224FeeManager`?
+
+### Target gas in node configuration
+
+`gas-target` in node configuration (json config) will be disabled upon activation of the `ACP176FeeManager` precompile. Should the precompile rather have a function to enable/disable that option? This will
```
michaelkaplan13, line 223: Looks like this section still needs to be completed.

michaelkaplan13 (commented):
Left a number of small comments. My primary bigger question is how changes to `q` are enforced for chains with the `ACP176FeeManager` precompile activated. 

Is it that each block must have a `q` value as specified by the latest call to `setFeeConfig` prior to the block? Or are there still incremental changes to `q` over the course of many blocks, and the `targetGas` specified in calls to `setFeeConfig` tells block builders what target gas they should be using in setting their preference? 

If it's the second, we'll also need a way to verify that `q` is incremented/decrement correctly according to the current value and latest precompile state I think.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: We probably want to fallback to C-chain defaults though (in case there is no config provided)

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: I'm not %100 sure but this might end up having to call `encode`/`decode` as follows (in external contract calls)
```
    function updateConfig(FeeConfig memory config) external {
        bytes memory encoded = abi.encode(config);
        feeConfigPrecompile.setFeeConfig(encoded);
    }
```

```
   function readConfig() external view returns (FeeConfig memory config) {
        bytes memory data = feeConfigPrecompile.getFeeConfig();
        config = abi.decode(data, (FeeConfig));
    }
```

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: the node config is `target-gas` flag in per-node configuration (as the one in c-chain). Are you implying to completely removing the node configuration and just depend on genesis chain config (or whatever the default is)? I was thinking to keep this mechanism in case some L1s (no owned subnets) want to leave this to their validators (like C-chain).

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: this is indeed a "open question" lol

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: Precompile activations itself is pretty much harmless, because we always use the fee config from the parent state. the only edge-case might be the deactivation of old precompile in the sense that how it should be done. i.e do we want to leave it to the configuratio, like expecting the configuration to disable old precompile first before activating new one. or else we can handle that in the activation of acp176 (force to disable the precompile). 

I added it as a open question.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
michaelkaplan13: That shouldn't be necessary, solidity handles the encoding/decoding of parameters for external function calls. 

This will be similar to how the ICM precompile function signatures looks actually: https://github.com/ava-labs/subnet-evm/blob/master/contracts/contracts/interfaces/IWarpMessenger.sol#L36

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
michaelkaplan13: Definitely agree that we want to keep this mechanism to allow for validators to set their preference, but we should require it to be specified on a per-chain basis for each node. I don't think there should be any `target-gas` node config that is applied to all EVM chains run by a given node. Each node should specify the `target-gas` for each chain individually, including the C-Chain.

ceyonur: ahh I see what you mean, no this should never be the case (and it would even more difficult to do so). What I mean by "node config" is a config that can be changed per node per chain. This will be equivalent to chainConfigs under `~/.avalanchego/configs/<chainID>/config.json`. I specifically wanted to avoid `chain config` because we also have a `chain config` in genesis json.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: right, I was under impression that we were using a config default.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Genesis Configuration
+
+There will be a new genesis chain configuration to set the parameters for the chain without requiring the ACP176FeeManager precompile to be activated. This will be similar to the existing fee configuration parameters in chain configuration. If there is no genesis configuration for the new fee parameters the default values for C-Chain will be used. This will look like the following:
+
+```json
+{
+  ...
+  "acp224Timestamp": ...,
+  "acp224FeeConfig": {
```
meaghanfitzgerald, line 111: The precompile interface is `IACP176FeeManager` but the genesis labels this as `acp224FeeConfig`. This will likely be confusing for users, ideally would have everything labeled with 176 since it defines the fee mechanism

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+### ACP-176 Parameters in Subnet-EVM
+
+ACP-176 will make `GasLimit` and `BaseFeeChangeDenominator` configurations obsolete in Subnet-EVM.
+
+`TargetBlockRate`, `MinBlockGasCost`, `MaxBlockGasCost`, and `BlockGasCostStep` will be kept same because ACP-176 still uses block gas cost to control the block production rate and surcharge for producing a block faster than the target rate. Subnet-EVM is configured to use following default values, which will be kept same:
+- `TargetBlockRate`: 2 (seconds)
+- `MinBlockGasCost`: 0 (gas)
+- `MaxBlockGasCost`: 1,000,000 (gas)
+- `BlockGasCostStep`: 200,000 (gas)
```
meaghanfitzgerald, line 76: Are these parameters formulaicly hardcoded or defining as hard coded variables somewhere? Understand that they are not included in the genesis config example but am unsure where or how these values could be returned

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,233 @@
...
+  "acp224Timestamp": ...,
+  "acp224FeeConfig": {
+    "targetGas": ...,
+    "minGasPrice": ...,
+    "maxCapacityFactor": ...,
+    "timeToDouble": ...,
+    "targetBlockRate": ...,
+    "minBlockGasCost": ...,
+    "maxBlockGasCost": ...,
+    "blockGasCostStep": ...
```
meaghanfitzgerald, line 119: should the precompile enforce any upper or lower bounds on these dynamic values?

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
ceyonur: I was going back and forth about using `ACP176` or `ACP224` mostly bc the `ACP176` defines the fee mechanism but `ACP224` defines the fee parameters for the mechanism. But I agree we should use common naming, and I guess `ACP176` makes more sense then?

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -0,0 +1,232 @@
...
+**BlockGasCostStep**:
+  Determines how much to increase/decrease the block gas cost depending on the amount of time elapsed since the previous block. If the block is produced at the target rate, the block gas cost will stay the same as the block gas cost for the parent block. If it is produced faster/slower, the block gas cost will be increased/decreased by the step value for each second faster/slower than the target block rate accordingly.
+  Note: if the `BlockGasCostStep` is set to a very large number, it effectively requires block production to go no faster than the `TargetBlockRate`.
+  Ex: if a block is produced two seconds faster than the target block rate, the block gas cost will increase by `2 * BlockGasCostStep`.
+
+### ACP-176 Parameters in Subnet-EVM
+
+ACP-176 will make `GasLimit` and `BaseFeeChangeDenominator` configurations obsolete in Subnet-EVM.
+
+`TargetBlockRate`, `MinBlockGasCost`, `MaxBlockGasCost`, and `BlockGasCostStep` will be also removed by [ACP-226](https://github.com/michaelkaplan13/ACPs/blob/main/ACPs/226-dynamic-minimum-block-times/README.md).
```
michaelkaplan13, line 71:
```suggestion
`TargetBlockRate`, `MinBlockGasCost`, `MaxBlockGasCost`, and `BlockGasCostStep` will be also removed by [ACP-226](https://github.com/avalanche-foundation/ACPs/tree/ce51dfab/ACPs/226-dynamic-minimum-block-times).
```

Lets link to the official repo (and use a permalink) here 👍

michaelkaplan13 (approved):

PR #226, opened 2025-08-05
Dynamic Minimum Block Times

michaelkaplan13:
Proposes replacing the current block production rate limiting mechanism on Avalanche EVM chains with a new mechanism where validators collectively and dynamically determine the minimum time between blocks.

### DIFF
--- a/ACPs/2xx-dynamic-minimum-block-times/README.md
+++ b/ACPs/2xx-dynamic-minimum-block-times/README.md
@@ -1,0 +1,119 @@
+| ACP | 2xx |
+| :- | :- |
+| **Title** | Dynamic Minimum Block Times |
+| **Author(s)** | Stephen Buttolph ([@StephenButtolph](https://github.com/StephenButtolph)), Michael Kaplan ([@michaelkaplan13](https://github.com/michaelkaplan13)) |
+| **Status** | Proposed, Implementable, Activated, Stale ([Discussion](POPULATED BY MAINTAINER, DO NOT SET)) |
+| **Track** | Standards |
+
+## Abstract
+
+Proposes replacing the current block production rate limiting mechanism on Avalanche EVM chains with a new mechanism where validators collectively and dynamically determine the minimum time between blocks.
+
+## Motivation
+
+Currently, Avalanche EVM chains employ a mechanism to limit the rate of block production by increasing the "block gas cost" that must be burned if blocks are produced more frequently than the target block rate specified for the chain. The block gas cost is paid by summing the "priority fee" amounts that all transactions included in the block collectively burn. This mechanism has a few notable suboptimal aspects:
+
+1. There is no explicit minimum block delay time. Validators are capable of producing blocks as frequently as they would like by paying the additional fee, and too rapid block production could cause network stability issues.
+1. The target block rate can only be changed in a required network upgrade, which makes updates difficult to coordinate and operationalize.
+1. The target block rate can only be specified with 1-second granularity, which does not allow for configuring sub-second block times as performance improvements are made to make them feasible.
+
+With the prospect of ACP-194 removing block execution from consensus and allowing for increases to the gas target through the dynamic ACP-176 mechanism, Avalanche EVM chains would be better suited by having a dynamic minimum block delay time denominated in milliseconds. This allows networks to ensure that blocks are never produced more frequently than the minimum block delay, and allows validators to dynamically influence the minimum block delay value by setting their preference.
+
+## Specification
+
+### Block Header Changes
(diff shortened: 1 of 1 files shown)


ACPs/226-dynamic-minimum-block-times/README.md
```diff
@@ -0,0 +1,123 @@
...
+1. The target block rate can only be specified with 1-second granularity, which does not allow for configuring sub-second block times as performance improvements are made to make them feasible.
+
+With the prospect of ACP-194 removing block execution from consensus and allowing for increases to the gas target through the dynamic ACP-176 mechanism, Avalanche EVM chains would be better suited by having a dynamic minimum block delay time denominated in milliseconds. This allows networks to ensure that blocks are never produced more frequently than the minimum block delay, and allows validators to dynamically influence the minimum block delay value by setting their preference.
+
+## Specification
+
+### Block Header Changes
+Upon activation of this ACP, the `blockGasCost` field in block headers will be required to be set to 0. This means that no validation of the cumulative priority fee amounts of transactions within the block exceeding the block gas cost is required. Additionally, two new fields will be added to EVM block headers: `timeMilliseconds` and `minimumBlockDelay`.
```
ARR4N, line 25: Both `uint64`?

ARR4N (approved):
Nice work, I'm supportive of this.

ACPs/226-dynamic-minimum-block-times/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+While this proposal requires a network upgrade and updates the EVM block header format, it does so in a way that tries to maintain as much backwards compatibility as possible. Specifically, applications that currently parse and use the existing timestamp field that is denominated in seconds can continue to do so. The `timeMilliseconds` header value only needs to be used in cases where more granular timestamps are required.
+
+## Reference Implementation
+
+A reference implementation is not yet provided, and must be made available for this ACP to be considered `implementable`.
+
+## Security Considerations
+
+Too rapid block production may cause availability issues if validators of the given blockchain are not able to keep up with blocks being proposed to consensus. This new mechanism allows validators to help influence the maximum frequency at which blocks are allowed to be produced, but potential misconfiguration or overly aggressive settings may cause problems for some validators.
```
yacovm, line 113: > Too rapid block production may cause availability issues if validators of the given blockchain are not able to keep up with blocks being proposed to consensus.

Can you elaborate on this in the text?

yacovm: By availability do you mean responsiveness for the end-user? Because the term is ambiguous, it can also refer to validator-to-validator availability.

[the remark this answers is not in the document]
yacovm: By halting the chain do you mean stopping block production? I know we have a [throttling check](https://github.com/ava-labs/avalanchego/blob/master/snow/engine/snowman/engine.go#L648) that prevents us from building blocks if we have too many of them waiting to be consented upon, but this is only temporary. 

I was mostly concerned about the terminology of availability issues as the association it brings me is nodes not responding to any kind of service (not only transaction inclusion but also any API call).

Perhaps it's better to say:

> Too rapid block production may impair block production, if ...

ACPs/226-dynamic-minimum-block-times/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Specification
+
+### Block Header Changes
+Upon activation of this ACP, the `blockGasCost` field in block headers will be required to be set to 0. This means that no validation of the cumulative priority fee amounts of transactions within the block exceeding the block gas cost is required. Additionally, two new fields will be added to EVM block headers: `timeMilliseconds` and `minimumBlockDelay`.
+
+#### `timeMilliseconds`
+
+The canonical serialization and interpretation of EVM blocks already contains a block timestamp specified in seconds. Altering this would require deep changes to the EVM codebase, as well as cause breaking changes to tooling such as indexers and block explorers. Instead, a new field is added representing the number of milliseconds past the existing block timestamp value denominated in seconds. The valid range of values for this field is 0 to 999. Wherever needed, the block timestamp in milliseconds can be easily constructed as:
+`(time * 1000) + timeMilliseconds`.
```
ceyonur, line 30: just to clarify
```suggestion
`(block.time * 1000) + block.timeMilliseconds`. 
```

ceyonur (commented):
I believe this will be a very welcome change for L1s. 👍

ACPs/226-dynamic-minimum-block-times/README.md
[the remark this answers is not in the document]
michaelkaplan13: > I was mostly concerned about the terminology of availability issues as the association it brings me is nodes not responding to any kind of service (not only transaction inclusion but also any API call).

I think that could also occur depending on the scenario, right? Nodes that aren't able to keep up with the chain because of too many blocks being proposed/accepted in consensus by other nodes could be marked as unhealthy and stop serving API requests, etc. That's why I left it as general "availability issues" really. If a node with 20+% of stake on a given L1 becomes unhealthy, the chain itself could have liveness problems from too much stake being offline.

PR #233, opened 2025-08-28
replace timeMilliseconds with timestampMilliseconds

ceyonur:
This PR replaces `timeMilliseconds` field (which only denotes ms part in the timestamp/0-999) with a full timestamp represented in milliseconds.


StephenButtolph (commented):
It may be obvious, but should we explicitly comment that block validity requires `block.timestampMilliseconds / 1000 == block.timestamp`?

PR #234, opened 2025-08-29
[ACP-226] minimumBlockDelayExcess and gas capacity added per millisecond.

michaelkaplan13:
This PR updates the ACP-226 specification with updates discussed on the community call on August 28th.

Specifically, the block headers will have `minimumBlockDelayExcess` field rather than `minimumBlockDelay`, and gas capacity will be added on a per millisecond basis to allow for sub-second block production even when a chain's gas capacity is fully consumed.

### DIFF
--- a/ACPs/226-dynamic-minimum-block-times/README.md
+++ b/ACPs/226-dynamic-minimum-block-times/README.md
@@ -39,7 +39,7 @@ The new `minimumBlockDelayExcess` field in the block header will be used to deri
 
 The `minimumBlockDelayExcess` field will be represented in block headers as a `uint64`.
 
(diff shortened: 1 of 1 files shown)


ACPs/226-dynamic-minimum-block-times/README.md
```diff
@@ -82,7 +88,7 @@ Parameters at activation on the C-Chain are:
 | Parameter | Description | C-Chain Configuration|
 | - | - | - |
 | $M$ | minimum `minimumBlockDelay` value | 100 milliseconds |
```
StephenButtolph, line 90: I think we should also make this 1ms to align with potential subnet-evm implementations.

StephenButtolph (approved):

PR #239, opened 2025-10-03
[ACP-181]: Spec Tweaks

michaelkaplan13:
- Specify epoch duration for Fuji and Mainnet.
- Update specification of epoch 0 start time definition.
- Align reference implementation with specification.
- Reference implementation cleanup.

### DIFF
--- a/ACPs/181-p-chain-epoched-views/README.md
+++ b/ACPs/181-p-chain-epoched-views/README.md
@@ -127,7 +127,7 @@ A full reference implementation that implements this ACP in the ProposerVM is av
 
 ### Setting the Epoch Duration
 
-The epoch duration $D$ is set on a network-wide level. For both Fuji (network ID 5) and Mainnet (network ID), $D$ will be set to 5 minutes upon activation of this ACP. Any changes to $D$ in the future would require another network upgrade.
+The epoch duration $D$ is set on a network-wide level. For both Fuji (network ID 5) and Mainnet (network ID 1), $D$ will be set to 5 minutes upon activation of this ACP. Any changes to $D$ in the future would require another network upgrade.
 
 #### Changing the Epoch Duration
 
--- a/ACPs/181-p-chain-epoched-views/README.md
+++ b/ACPs/181-p-chain-epoched-views/README.md
@@ -30,7 +30,7 @@ An epoch is defined as a contiguous range of blocks that share the same three va
 
 Let $E_N$ denote an epoch with epoch number $N$. $E_N$'s start time is denoted as $T_{start}^N$, and its P-Chain height as $P_N$. 
 
-$E_0$ is defined as the epoch whose start time $T_{start}^0$ is the block timestamp of the parent block of the block that activates this ACP (i.e. the last block prior to this ACP's activation timestamp).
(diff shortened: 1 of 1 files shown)


ACPs/181-p-chain-epoched-views/README.md
```diff
@@ -101,7 +101,7 @@ type Block interface {
 }
 
 func GetPChainEpoch(parent Block) Epoch {
```
StephenButtolph, line 103: Perhaps out of scope of this fix, But I think it would be cleaner if we did:
```go
func GetPChainEpoch(parent Block) Epoch {
	parentTimestamp := parent.Timestamp()
	parentEpoch := parent.Epoch()
	epochEndTime := parentEpoch.StartTime.Add(D)
	if parentTimestamp.Before(epochEndTime) {
		// If the parent was issued before the end of its epoch, then it did not
		// seal the epoch.
		return parentEpoch
	}

	// The parent sealed the epoch, so the child is the first block of the new
	// epoch.
	return Epoch{
		PChainHeight: parent.PChainHeight(),
		Number:       parentEpoch.Number + 1,
		StartTime:    parentTimestamp,
	}
}
```

Essentially, I think `Before` is easier to read/understand than `not Before`. Also I just find the comments overly verbose... It just says what the code does.

ACPs/181-p-chain-epoched-views/README.md
```diff
@@ -30,7 +30,7 @@ An epoch is defined as a contiguous range of blocks that share the same three va
 
 Let $E_N$ denote an epoch with epoch number $N$. $E_N$'s start time is denoted as $T_{start}^N$, and its P-Chain height as $P_N$. 
 
-$E_0$ is defined as the epoch whose start time $T_{start}^0$ is the block timestamp of the block that activates this ACP (i.e. the first block at or following this ACP's activation timestamp).
+$E_0$ is defined as the epoch whose start time $T_{start}^0$ is the block timestamp of the parent block of the block that activates this ACP (i.e. the last block prior to this ACP's activation timestamp).
```
StephenButtolph, line 33: Should we also comment on what the initial pChainHeight is here?

StephenButtolph (approved):

michaelkaplan13:
> nsiderations prompted the change to the initial epoch def

michaelkaplan13:
> LGTM! Just curious what considerations prompted the change to the initial epoch definition? Since a block's inclusion in an epoch is inclusive of the start time, whether the initial epoch starts at the activating block's timestamp or its parent seems like it should be identical, but there may be an edge case I'm not considering.

IMO, it could have gone either way, but this definition makes it such that the first epoch is not a special case.

An epoch's start time must be the timestamp of the block that sealed the previous epoch in order for the epoch to be able to advance with every block (i.e. if every block is at least `D` seconds apart, every block is it's own epoch). And an epoch's P-Chain height needs to be the latest P-Chain height of the block that sealed the previous epoch such that the epoched height of the next block is always known before it's produced. 

The last block before activation doesn't seal any epoch, but it does have a timestamp and P-Chain available to use, so this definition best aligns the first epoch with the common case.

PR #255, opened 2025-12-11
[ACP-255] New Curve Model for L1 Validator PAYG Fee

iJaack:
This PR proposes a new Gaussian bell curve fee model for L1 validators, replacing ACP-77's flat-fee structure. Combined with L1-size-dependent multipliers, it dramatically increases AVAX burn while incentivizing validator decentralization.

## Problem

Current ACP-77 model:
- **Insufficient burn:** Only 1,064 AVAX/month at 800 validators
- **No decentralization incentive:** Flat fees favor single-validator L1s
- **Misaligned growth:** Network expansion provides no economic benefit

## Solution

**L1-Size Multiplier:** Cost plateau at 10-15 validators makes validator set growth economically irresistible (15-validator L1 costs same as 1-validator, but 15x more secure).

**Gaussian Network Factor:** Bell curve centered at 10,000 validators creates time-dependent incentives—high burn during bootstrap, peak at critical scale, graceful decay as network matures.

## Expected AVAX Burn Rates

| Network | Validators | Current | Proposed | Increase |
|---------|-----------|---------|----------|----------|
| **Current** | **800** | **1,064** | **9,040** | **8.5x** |
| Growth | 5,000 | 6,650 | 97,800 | 14.7x |
| **PEAK** | **10,000** | **13,300** | **266,200** | **20.8x** |
| Sustainable | 50,000 | 66,500 | 347,000 | 5.2x |

(AVAX/month burned)

## Benefits

- **Deflation:** 8.5-20.8x higher AVAX burn strengthens long-term value
- **Decentralization:** Cost plateau eliminates incentive for single-validator L1s
- **Growth Alignment:** Network expansion reduces per-validator costs, creating adoption spiral
- **Elegance:** Continuous smooth curve, no discontinuities

## Implementation

- Base rate doubled (512 → 1,024 nAVAX/sec)
- L1 multiplier: 1.0x to 18.84x based on validator count
- Network factor peaks at 3.84x at 10,000 validators

_(Sorry for the commit mix-up, I created the ACP-247 PR in my main branch, I now have separate branches for each ACP)_

### DIFF
--- a/ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
+++ b/ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
@@ -204,6 +204,55 @@ ACP-255's cost plateau at 10-15 validators **aligns with the natural maturity ta
 
 **Economic equilibrium:** The majority of L1s will naturally converge to 10-15 validators over 12-24 months, driven by cost optimization rather than mandates.
 
+## Canonical Scenario Tables
+
+To make the internal-margin and external-margin effects easier to compare, this section reduces the proposal to three explicit scenario paths under the same assumptions.
+
+### Scenario A: Enterprise L1 Lifecycle (3 -> 5 -> 8 validators)
+
+This scenario models an enterprise-style L1 that values operator control and grows conservatively.
+
+| Stage | Network Validators | L1 Validators | ACP-77 Total Monthly Cost | ACP-255 Total Monthly Cost | ACP-255 Per-Validator Cost | Interpretation |
+|---|---|---|---|---|---|---|
+| Launch | 800 | 3 | 3.98 AVAX | 140.12 AVAX | 46.71 AVAX | Strong centralization premium at launch |
+| Early growth | 2,000 | 5 | 6.64 AVAX | 161.57 AVAX | 32.31 AVAX | Total cost rises modestly, but per-validator cost falls materially |
+| Mature enterprise set | 5,000 | 8 | 10.62 AVAX | 190.76 AVAX | 23.84 AVAX | L1 remains expensive, but growth reduces the centralization penalty |
+
+**Internal-margin takeaway:** Under ACP-77, the enterprise L1 is always incentivized to stay as small as possible. Under ACP-255, it may still rationally remain below the 10-15 validator plateau, but every added validator meaningfully reduces per-validator cost. In other words, ACP-255 does not forbid smaller enterprise sets, it taxes them.
+
+### Scenario B: Community L1 Lifecycle (3 -> 10 -> 15 validators)
+
+This scenario models a community-oriented L1 that grows toward the 10-15 validator plateau.
+
+| Stage | Network Validators | L1 Validators | ACP-77 Total Monthly Cost | ACP-255 Total Monthly Cost | ACP-255 Per-Validator Cost | Interpretation |
+|---|---|---|---|---|---|---|
+| Launch | 800 | 3 | 3.98 AVAX | 140.12 AVAX | 46.71 AVAX | Small validator sets are deliberately expensive |
+| Growth plateau entry | 2,000 | 10 | 13.27 AVAX | 111.49 AVAX | 11.15 AVAX | More validators but lower total cost than the 3-validator launch case |
+| Mature community set | 5,000 | 15 | 19.91 AVAX | 142.36 AVAX | 9.49 AVAX | Plateau behavior appears: per-validator cost keeps falling while total cost stays manageable |
+
+**Internal-margin takeaway:** This is the core behavioral target of ACP-255. A community L1 that expands from 3 to 10 validators becomes much more decentralized while also lowering its total fee burden relative to its own launch state under ACP-255. The proposal therefore changes validator growth from a pure cost center into a path toward fee efficiency.
+
+### Scenario C: New Entrant Launch Decision (1 vs 3 vs 10 validators)
+
+This scenario isolates the external-margin decision for a brand-new L1 entering a network with 800 total validators.
+
+| Launch Configuration | Network Validators | ACP-77 Total Monthly Cost | ACP-255 Total Monthly Cost | ACP-255 Per-Validator Cost | Interpretation |
+|---|---|---|---|---|---|
+| 1 validator | 800 | 1.33 AVAX | 81.54 AVAX | 81.54 AVAX | Single-validator entry becomes economically unattractive |
+| 3 validators | 800 | 3.98 AVAX | 140.12 AVAX | 46.71 AVAX | Small launches are possible, but pay a steep centralization premium |
+| 10 validators | 800 | 13.27 AVAX | 95.18 AVAX | 9.52 AVAX | Launching closer to the plateau is far more cost-efficient on a per-validator basis |
+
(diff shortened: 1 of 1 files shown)


ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
```diff
@@ -0,0 +1,242 @@
...
+- Current network (800 validators): **8.5x higher burn** (1,064 → 9,040 AVAX/month)
+- At 2,000 validators: **9.9x higher burn** (2,660 → 26,360 AVAX/month)
+- At 5,000 validators: **14.7x higher burn** (6,650 → 97,800 AVAX/month)
+- **At 10,000 validators: 20.8x higher burn (PEAK)** (13,300 → 266,200 AVAX/month)
+- At 20,000 validators: **8.2x higher burn** (26,600 → 218,000 AVAX/month)
+- At 50,000 validators: **5.2x higher burn** (66,500 → 347,000 AVAX/month)
+
+## Motivation
+
+The current ACP-77 flat-fee model provides insufficient burn (~1,064 AVAX/month) and creates no economic incentive for validator decentralization. ACP-255 implements a smooth Gaussian fee curve that:
```
michaelkaplan13: The current PAYG fee mechanism is actually not a flat-fee model. It is a dynamic fee model with a minimum fee of 512 nAVAX/sec, and a target of 10,000 L1 validators. When the target is exceeded, the fee rate begins to rise exponentially until the total number of validators returns to the target. When the total number of validators is beneath the target, the fee rate decreases down to the minimum.

(specified by ACP-77 [here](https://github.com/iJaack/ACPs/tree/acp-255/ACPs/255-new-curve-model-for-l1-validator-payg-fee))

ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
```diff
@@ -0,0 +1,242 @@
...
+2. **L1-Size-Dependent Fee Multiplier:** Maximum premium of 50 AVAX/month for single-validator L1s, creating powerful decentralization incentives
+3. **Gaussian Network Factor:** Elegant bell curve centered at 10,000 validators with 3.84x factor peak
+
+**Economic Impact:**
+- Current network (800 validators): **8.5x higher burn** (1,064 → 9,040 AVAX/month)
+- At 2,000 validators: **9.9x higher burn** (2,660 → 26,360 AVAX/month)
+- At 5,000 validators: **14.7x higher burn** (6,650 → 97,800 AVAX/month)
+- **At 10,000 validators: 20.8x higher burn (PEAK)** (13,300 → 266,200 AVAX/month)
+- At 20,000 validators: **8.2x higher burn** (26,600 → 218,000 AVAX/month)
+- At 50,000 validators: **5.2x higher burn** (66,500 → 347,000 AVAX/month)
```
michaelkaplan13: The maximum number of L1 validators is 2x the target of 10,000, so there can not be anymore than 20,000 in total currently.

ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+**Fee Schedule:**
+
+| L1 Validators | Multiplier | Fee/Validator | L1 Total |
+|---|---|---|---|
+| 50+ | 1.00x | 2.65 AVAX | Variable |
+| 15 | 1.27x | 3.36 AVAX | 50.46 AVAX |
+| **10** | **2.20x** | **5.84 AVAX** | **58.36 AVAX** |
+| 5 | 6.37x | 16.91 AVAX | 84.57 AVAX |
+| 1 | 18.84x | 50.00 AVAX | 50.00 AVAX |
```
michaelkaplan13: Unfortunately, this also present a DOS vulnerability for the P-Chain and primary network as a whole.

Each additional L1 validator uses up more resources for primary network validators to track, and thus _must_ result in additional fees being paid. 

I like the idea of each validator paying less the more validators there are, but it must be the case that in aggregate, more fees are paid for each additional validator.

ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
iJaack: Oh OK, is there a documentation link to support that? Would love to dive into it.

ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
iJaack: @ericlu-avax in this new model, the total fees paid by a chain would be the same, but instead of just being dependent on global validator count, it would also be dependent on the single validator set.

Does that make sense? Does this answer the question?

iJaack:
Just cleaned the commit list for this PR to ensure it's not mixed with other ACPs.

Would love to get more feedback as to if this proposal is doable, maybe in multiple steps.

cc @michaelkaplan13 @ericlu-avax

iJaack:
@ericlu-avax what do you think about the feasibility of this ACP? I know this is way harder and complex to implement, mostly because it includes changing not just some variables, but how those variables work and even add other variables on top.

Would love to discuss it in a community call cc @meaghanfitzgerald

ericlu-avax:
> @ericlu-avax what do you think about the feasibility of this ACP? I know this is way harder and complex to implement, mostly because it includes changing not just some variables, but how those variables work and even add other variables on top.
> 
> Would love to discuss it in a community call cc @meaghanfitzgerald

Hi @iJaack, the complexity to implement might be better directed towards the engineering team. The missing piece imo is still the same as before: in order to evaluate the expected impact on the L1 fee generation on both the internal and external margin, we need to compare the current and proposed mechanisms under the same set of assumptions mainly on the evolution of L1s over time. For internal margin, we want to  know how would fee process change for an existing L1 with its expected growth trajectory. For external margin, we want to know how would a potential L1's entry decision change given the different fee profiles. It is not clear to me what would be the reasonable set of assumptions to get at these questions.

iJaack:
Thanks @ericlu-avax , this helps clarify what the missing piece is.

I have now added sections to the ACP that try to answer this in the exact frame you described:

- **Core Formulas**: consolidates the full fee functions and parameters
- **L1 Validator Growth Assumptions**: makes the lifecycle assumptions explicit for enterprise vs community L1s, plus an expected network growth path
- **Marginal Cost Comparison**: compares the current and proposed mechanisms under the same validator-count assumptions, including the cost of adding validators at different L1 sizes
- **Game-Theoretic Equilibrium**: frames the internal and external margin more directly, i.e. how an existing L1's incentives evolve as it grows, and how the entry / optimal-size decision changes for a new L1
- **Interaction with ACP-247**: explains how the fee side and reward side can be evaluated together rather than in isolation

iJaack:
I added the scenario tables directly to the ACP now.

New section: **Canonical Scenario Tables**
- Scenario A: enterprise L1 lifecycle `3 -> 5 -> 8`
- Scenario B: community L1 lifecycle `3 -> 10 -> 15`
- Scenario C: new entrant launch decision `1 vs 3 vs 10`

The goal was to make the internal-margin and external-margin comparison explicit under the same assumptions, rather than leaving it implicit across the other sections.

Nuttymoon:
> I added the scenario tables directly to the ACP now.

An L1 should realistically never have fewer than 5 validators. If a single node goes down, it'd take the whole network down with it. That's why AvaCloud and other service providers start at 5 validators.

PR #265, opened 2026-01-21
[ACP-265] Interchain Non-Fungible Token Transfer (ICNFTT)

midnight-commit:
This specification defines ICNFTT (Interchain Non-Fungible Token Transfer), a best practices standard for NFT (ERC721) transfers between Avalanche L1s using Interchain Messaging (ICM) and Teleporter.


iJaack:
Hey, first of all this is ACP-265, based on issue numbering :)

Then, I guess you opened this ACP because you either need this or you think someone needs or will need this. Do you have some strong motivations / real world use cases?

midnight-commit:
Hi @iJaack,

fixed the ACP number, wasn't sure what it's based on tbh, now I know, thank you. 

We developed this for a client in cooperation with Ava Labs. The main idea behind it is to allow cheap mass minting of NFTs on your custom Avalanche L1 and then allow users to transfer select items to C-Chain, which all the NFT marketplaces integrate. Or to bring existing NFT collections to your L1 where they can be used in a novel way.

PR #269, opened 2026-02-08
[ACP-224] Add StaticPricing and ValidatorTargetGas modes 

ceyonur:
Added new 2 modes: 
`StaticPricing`: to specify a fixed price at min gas 
`ValidatorTargetGas`: added ability to precompile to give the gas target preference back to validators.

Also made various clarifications and cleanups.  

Note: For some reason I cannot edit the title


ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -138,6 +145,23 @@ interface IACP224FeeManager is IAllowList {
...
+
+    /// @notice Emitted when validator target gas mode is enabled
+    /// @param sender Address that triggered the change
+    event ValidatorTargetGasEnabled(address indexed sender);
+
+    /// @notice Emitted when validator target gas mode is disabled
+    /// @param sender Address that triggered the change
+    /// @param targetGas The new authoritative target gas value
+    event ValidatorTargetGasDisabled(address indexed sender, uint256 targetGas);
+
```
michaelkaplan13: Would it be more standard to a single event for each of these, with a boolean parameter for whether or not it was enabled/disabled?

The same applies for the external functions as well. I think (for example) `setStaticPricing(bool enabled)` likely clean the interface up a bit.

michaelkaplan13: Actually...could we just add these two fields to the `FeeConfig` struct, and then leave the rest of the interface as is? 

For consistency, we should either have separate functions to set each parameter of the configuration, or one function to set them all.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -151,15 +162,29 @@ interface IACP224FeeManager is IAllowList {
...
+**Toggling behavior:**
+
+- **Enabling** (setting `validatorTargetGas` to `true` via `setFeeConfig`): Validators begin adjusting `targetExcess` from the parent block's current value. Since ACP-176's bounded update mechanism limits changes to roughly $\frac{1}{1024}$ of the current value per block, the transition is always gradual. The parent block's `targetExcess` (which was set by the precompile) provides an unambiguous starting point.
+- **Disabling** (setting `validatorTargetGas` to `false` via `setFeeConfig`): The `targetGas` value provided in the same `setFeeConfig` call is stored and immediately becomes authoritative for block building. The caller should check the current effective gas target (from block headers or RPC) and provide an appropriate `targetGas` value, since validators may have drifted the effective gas target far from the previously stored value and using a stale value could cause a dangerous sudden jump in gas capacity.
```
michaelkaplan13: These changes only take effect upon settlement of the transaction that makes the call to the precompile, right?
```suggestion
- **Disabling** (setting `validatorTargetGas` to `false` via `setFeeConfig`): The `targetGas` value provided in the same `setFeeConfig` call is stored and immediately authoritative for block building upon settlement. The caller should check the current effective gas target (from block headers or RPC) and provide an appropriate `targetGas` value, since validators may have drifted the effective gas target far from the previously stored value and using a stale value could cause a dangerous sudden jump in gas capacity.
```

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
```diff
@@ -171,19 +196,23 @@ The precompile configuration will look like the following:
...
   }
 }
 ```
 
 The `initialFeeConfig` fields are:
 
-- `targetGas` (uint64, **required**): Target gas consumption per second ($T$).
-- `minGasPrice` (uint64, **required**): Minimum gas price in wei ($M$).
-- `timeToDouble` (uint64, **required**): Seconds for the gas price to double when the chain is at maximum capacity. Determines $K$ as described in [ACP-176 Parameters in Subnet-EVM](#acp-176-parameters-in-subnet-evm).
+- `validatorTargetGas` (bool, optional, default `false`): When `true`, validators control `targetGas` dynamically via their node preferences, using the same mechanism as the C-Chain (see [Dynamic Gas Target Via Validator Preference](#dynamic-gas-target-via-validator-preference)). The stored `targetGas` value is not used for block building; validators always adjust from the parent block's `targetExcess`.
```
michaelkaplan13: A good defensive check here would be to require that `targetGas` equal 0 whenever `validatorTargetGas` is `true`. If `validatorTargetGas` is `false`, I believe `targetGas` has a minimum of 1,000,000.

Similarly, if `staticPricing` is `true` we could require that `timeToDouble` is 0.

ACPs/224-dynamic-gas-limit-in-subnet-evm/README.md
michaelkaplan13: Not new in this PR, but this doesn't display properly in github: https://github.com/ceyonur/ACPs/tree/ceyonur/acp-224-revisit/ACPs/224-dynamic-gas-limit-in-subnet-evm#internal-state

michaelkaplan13: This also isn't new in this PR, but isn't display correctly: https://github.com/ceyonur/ACPs/tree/ceyonur/acp-224-revisit/ACPs/224-dynamic-gas-limit-in-subnet-evm#adjustment-to-acp-176-calculations-for-price-discovery

michaelkaplan13 (approved):

avalanche-foundation-admin (approved):

PR #273, opened 2026-02-12
[ACP-273] Reduce Minimum Validator Staking Period to 24 Hours

meaghanfitzgerald:
This proposal reduces the minimum validator staking period on Avalanche's Primary Network from 2 weeks (336 hours) to 1 day (24 hours). The change lowers barriers to validator participation while maintaining network security, enabling more flexible staking strategies and improving capital efficiency across the ecosystem.


ACPs/273-24-hour-minimum-staking-period/README.md
```diff
@@ -0,0 +1,66 @@
...
+### Implementation Details
+
+The staking mechanism remains unchanged. Validators must still meet all existing requirements:
+
+- Minimum stake: 2,000 AVAX for Primary Network validators
+- Uptime requirement: 90% (per ACP-267)
+- Hardware requirements: As specified in ACP-256
+
+Only the minimum duration parameter changes, with all other validation rules and reward calculations preserved.
+
```
martineckardt: Add Open Questions section with the point if the min duration of delegations should also be lowered, potentially with requiring a larger min delegation amount (e.g 1,000 AVAX) for shorter delegations.

martineckardt (changes requested):

ACPs/273-24-hour-minimum-staking-period/README.md
```diff
@@ -0,0 +1,74 @@
...
+### Benefits of a Shorter Minimum Period
+
+1. Increased Validator Participation: A 24-hour minimum removes a significant barrier to entry. Validators who cannot commit capital for 2 weeks can now participate, increasing overall network stake and decentralization.
+
+2. Improved Capital Efficiency: Stakers gain reliable access to liquidity after 24 hours rather than 14 days. This enables more dynamic capital allocation strategies while maintaining commitment to network security during the staking period.
+
+### Reward Impact Analysis
+
+Using Avalanche's reward formula at the time of writing, the APY difference between 24-hour and 2-week staking periods is approximately 0.04 percentage points (6.08% vs 6.12%). This minimal difference preserves reward incentives while enabling liquidity flexibility.
```
michaelkaplan13, line 26: This is more so for discussion after this PR gets merged, but IMO this is _way_ too small of an incentive to lock for two weeks rather than for 1 day.

I think the question here is along the lines of what is the "cost" of losing access to liquidity for the longer period of time, because the incentive to do so must more than offset it in order for it to be effective.

Shorter staking durations provide lower security to the network as a whole, so its important that the incentive to staking for longer be strong enough for people to do so.

ACPs/273-24-hour-minimum-staking-period/README.md
```diff
@@ -0,0 +1,74 @@
...
+- Uptime requirement: 90% (per ACP-267)
+- Hardware requirements: As specified in ACP-256
+
+Only the minimum duration parameter changes, with all other validation rules and reward calculations preserved.
+
+## Backwards Compatibility
+
+This change affects only new staking periods initiated after activation. Existing validators with active staking periods remain unaffected and will complete their original durations under the previous rules.
+
+The change is non-breaking: all existing validation infrastructure, reward mechanisms, and consensus operations continue functioning without modification.
```
michaelkaplan13, line 60: It should be mentioned here that this change requires a network upgrade to be implemented since it is a non-backwards compatible change to the P-Chain validation/execution requirements.

ACPs/273-24-hour-minimum-staking-period/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Security Considerations
+
+Validators remain subject to the same accountability standards during the 24-hour period. Network consensus sampling assumes the same validator availability model. Historical data demonstrates that validator uptime patterns remain consistent regardless of staking duration length, as infrastructure quality and operational commitment drive uptime, not duration requirements alone.
+
+## Open Questions
+
+1. If the minimum validation period is shortened, should the minimum delegation period also be reduced?
+
+At the time of writing, the ratio of AddPermissionlessDelegatorTxs to AddPermissionlessValidatorTxs over the past 365 days was 100:1, with delegations accounting for 46% of all P-Chain transactions. Assuming all delegations are currently set to the minimum duration (14 days), reducing the minimum to 24 hours would increase the rate of state growth from delegation operations on the P-Chain by approximately 14x.
```
michaelkaplan13:
```suggestion
At the time of writing, the ratio of `AddPermissionlessDelegatorTx`s to `AddPermissionlessValidatorTx`s over the past 365 days was 100:1, with delegations accounting for 46% of all P-Chain transactions. Assuming all delegations are currently set to the minimum duration (14 days), reducing the minimum to 24 hours would increase the rate of state growth from delegation operations on the P-Chain by approximately 14x.
```

ACPs/273-24-hour-minimum-staking-period/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+Only the minimum duration parameter changes, with all other validation rules and reward calculations preserved.
+
+## Backwards Compatibility
+
+This change affects only new staking periods initiated after activation. Existing validators with active staking periods remain unaffected and will complete their original durations under the previous rules.
+
+The change is non-breaking: all existing validation infrastructure, reward mechanisms, and consensus operations continue functioning without modification.
+
+## Security Considerations
```
michaelkaplan13, line 62: Not sure if it should go under "Security Considerations" or "Open Questions" but it also needs to be considered that shorter staking periods increase the likelihood of significant validator churn over a short period of time. This potentially impacts consensus safety and the ability to deliver ICM messages.

We should also add an open question on the difference in rewards rate for shorter vs longer validation periods. As an example of the significant security risk of short term staking periods, consider the case where the incentive to have a >24 staking duration is insufficient, and the vast majority of stake opts for a 24 hour duration (particularly likely with the addition of ACP-236). Then it would be the case that within 24 hours, basically all of the entities securing the network can be changed completely. The entire validator set could disappear, or a 1% validator could within a day become a 20% validator. This instability has a very real cost in terms of network security.

ACPs/273-24-hour-minimum-staking-period/README.md
meaghanfitzgerald: Agreed, in the future the avalanche foundation should evaluate the reward curve formula and reassess to capture that "cost" in the rewards calculated.

ACPs/273-24-hour-minimum-staking-period/README.md
meaghanfitzgerald: Added in line 56. Could you or @StephenButtolph confirm that line 58 is accurate? 

`This change affects only new staking periods initiated after activation. Existing validators with active staking periods remain unaffected and will complete their original durations under the previous rules.`

ACPs/273-24-hour-minimum-staking-period/README.md
meaghanfitzgerald: Added both of these points to the security considerations, and included a question on how can we incentive longer staking periods to the open questions section. Ideally, this question can be answered and implemented in the same upgrade as this acp to ensure the security of the network.

michaelkaplan13 (approved):

avalanche-foundation-admin (approved):

PR #275, opened 2026-02-18
ACP-275: L1 Trust Profile Framework

iJaack:
## Summary

This ACP proposes a standardised, neutral methodology for describing the trust characteristics of Avalanche L1 blockchains across four independent dimensions:

1. **Validator Control** — who decides validator entry and exit
2. **Validator Distribution & Bridged Asset Security** — how concentrated validator operation is, and what that means for bridged asset risk under ICM's 67% BLS threshold
3. **Censorship Resistance** — whether a controlling party can selectively exclude or delay transactions
4. **Governance Transparency** — how visible and predictable upgrade processes are

Each dimension is scored using four labels — **SP / MP / MS / OP** (Single-Party, Multi-Party, Multi-Stakeholder, Open) — all describing the same axis: who holds effective authority over that dimension, from most concentrated to fully open.

The framework produces a 4-dimensional profile per L1. No composite score, no ranking, no prescription. Different L1s have different goals; the framework describes, it does not judge.

## Board & Process

Governance and scoring is coordinated by an **L1 Trust Profile Board** (5–7 members), with explicit conflict-of-interest disclosure rather than exclusion. A bootstrapping process is defined: the Avalanche Foundation nominates the initial Board within 30 days of adoption, with community ratification via Snapshot within 60 days.

Reviews are both **annual** (full re-assessment of all active L1s) and **change-triggered** (expedited within 30 days when an L1 makes a material change to validator policy, operator composition, bridging, or governance).

## Track Note

This ACP is filed as **Best Practices** to reflect its non-protocol-changing, informational nature. Given its methodological and procedural scope, the community may wish to consider whether Standards Track is more appropriate.

## Credits
_Thanks to the AVAX Economics Working Group, especially to @Nuttymoon and @selim5335_

### DIFF
--- a/ACPs/274-l1-stages-classification-framework/README.md
+++ b/ACPs/274-l1-stages-classification-framework/README.md
@@ -1,26 +1,28 @@
 | ACP | 274 |
 |:--------------|:------------------------------------------------------------|
-| **Title** | L1 Stages Classification Framework |
+| **Title** | L1 Trust Profile Framework |
 | **Author(s)** | Giacomo Barbieri ([@ijaack94](https://x.com/ijaack94)) |
-| **Status** | Proposed  ([Discussion]()) |
+| **Status** | Proposed ([Discussion]()) |
 | **Track** | Best Practices |
+
+<!-- Track note: This ACP is filed as Best Practices to reflect its non-protocol-changing, informational nature. However, given that it defines a shared methodology, a Board, and a classification process intended for broad ecosystem adoption, the community may wish to consider whether Standards Track is more appropriate. -->
 
 ## Abstract
 
-The L1 Stages Classification Framework is a standardized, neutral methodology for describing the trust characteristics of Avalanche Layer 1 (L1) blockchains. As the Avalanche L1 ecosystem expands, participants require clear, consistent, and transparent information about validator control, operator diversity, bridged asset security, and governance transparency.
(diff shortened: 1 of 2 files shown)


iJaack:
Sorry, I was 100% sure that this had to be ACP-274 but it seems like I got it wrong, so the name of the ACP is still 275, but the branch is `acp-274`..

ACPs/274-l1-stages-classification-framework/README.md
```diff
@@ -0,0 +1,496 @@
...
+4. Governance Transparency
+
+Each dimension is scored using one of four labels:
+
+| Label | Full Name | Controlling Party |
+|---|---|---|
+| **SP** | Single-Party | One entity |
+| **MP** | Multi-Party | Small known group (2–5) |
+| **MS** | Multi-Stakeholder | Broad community / many parties |
+| **OP** | Open | No human gating; protocol-determined |
```
Nuttymoon, line 93: IMO there is a problem with the labels:
- The difference between `Muti-Party` and `Multi-Stakeholder` looks subjective. It is complicated to reason about it if the threshold depends on each dimension
- A network could be `Open` and yet be `SP`, `MP`, or `MS` at a given point

ACPs/255-new-curve-model-for-l1-validator-payg-fee/README.md
Nuttymoon, line 1: This file should probably not be here

Nuttymoon (commented):
As I have [expressed in a previous discussion](https://github.com/avalanche-foundation/ACPs/discussions/249#discussioncomment-15015386), my stance is that having L1 stages and a logical progression between them would be best.  
I also fear that the 4x4 profile proposed would be hard to reason about for both L1 teams and users.

At Suzaku, we've been working internally on gathering data about all L1s and classifying them following the criteria and stages I laid out in the comment linked above. We'll share the results soon.

PR #285, opened 2026-04-14
[ACP-285] Reduce Minimum Consumption Rate 

meaghanfitzgerald:
Proposes reduces MinConsumptionRate from 10% to 7.5% in Primary Network genesis, widening the APY spread between minimum-duration and maximum-duration staking from 1.02 pp to 2.30 pp without changing the maximum reward for 365-day validators. Addresses ACP-273 open question about whether rewards should differ by duration to incentivize network stability and provides the incentive gradient needed before ACP-273 + ACP-236 activate. 

Single constant change across three genesis files; no modifications to reward formula logic.


ACPs/285-reduce-minimum-consumption-rate/README.md
```diff
@@ -0,0 +1,150 @@
...
+| Minimum validator stake | 2,000 AVAX |
+| Minimum delegator stake | 25 AVAX |
+| Uptime requirement | 90% (per [ACP-267](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/267-uptime-requirement-increase/README.md)) |
+| Reward formula structure | Unchanged |
+
+## Backwards Compatibility
+
+This is a non-backwards compatible change to P-Chain reward calculation. It requires a network upgrade to activate.
+
+The change affects only new staking periods initiated after activation. Validators and delegators with active staking positions at the time of activation continue to earn rewards under the original parameters for the remainder of their existing stake periods. No migration or re-registration is required.
```
meaghanfitzgerald: @StephenButtolph is this true? or will all rewards be affected post activation?

ACPs/285-reduce-minimum-consumption-rate/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Reducing `MinConsumptionRate` from 10% to 7.5% produces the following APY schedule:
+
+| Duration | APY (MinRate = 7.5%) | Compounding APY (MinRate) = 7.5% |
+|----------|----------------------|----------------------------------|
+| 2 days   | 4.00%                | 4.08%                            |
+| 14 days  | 4.08%                | 4.16%                            |
+| 90 days  | 4.58%                | 4.65%                            |
+| 182 days | 5.18%                | 5.25%                            |
+| 365 days | 6.38%                | 6.38%                            |
```
Nuttymoon: I'd add the current APR here for clarity.  
Also, APY is misused. It should be APR if not taking compounding into account. (The same goes for the rest of the ACP)
```suggestion
| Duration | Current APR (MinRate = 10%) | APR (MinRate = 7.5%) | Compounding APY (MinRate) = 7.5% |
| -------- | --------------------------- | -------------------- | -------------------------------- |
| 2 days   | -                           | 4.00%                | 4.08%                            |
| 14 days  | 5.36%                       | 4.08%                | 4.16%                            |
| 90 days  | 5.58%                       | 4.58%                | 4.65%                            |
| 182 days | 5.85%                       | 5.18%                | 5.25%                            |
| 365 days | 6.38%                       | 6.38%                | 6.38%                            |
```

hinkavax:
## Response to ACP-285: A Targeted Alternative That Preserves Existing Staker Economics

### Summary

We support the goal of ACP-285 — steepening the reward gradient to discourage ultra-short rolling stakes enabled by ACP-273 + ACP-236. The security concern is valid and we agree this should co-activate with ACP-273.

However, the proposed mechanism — a blanket reduction of `MinConsumptionRate` — is too blunt. It solves a forward-looking problem created by ACP-273 by retroactively penalizing existing staker behavior that has been securing the network without incident since mainnet launch.

We propose a piecewise linear alternative that achieves a comparable incentive gradient for ultra-short stakers while leaving all existing staker economics completely unchanged.

---

### The 1.02% Spread Isn't Broken 

ACP-285 frames the current 1.02% spread between 14-day and 365-day stakers as evidence of weak duration incentives. However, 1.02% is only "small" in absolute terms.** On a base of 5.36%, an additional 1.02% represents a 19% relative increase in yield. If a bank offered 19% more return for locking a term deposit, that is a meaningful premium. For context, the US Treasury curve spread between 1-month and 12-month bills frequently sits below 100bp. Investors think in relative terms.

---

### Why the Current Proposal Creates Collateral Damage

The linear reward formula places 48-hour and 14-day stakers in nearly the same position on the curve (0.5% vs 3.8% of the minting period). Lowering `MinConsumptionRate` penalizes both almost equally:

| Duration | Current APR | ACP-285 APR | Change |
|----------|------------|-------------|--------|
| 48 hours | - | - | - |
| 14 days | 5.36% | 4.08% | -1.28pp |
| 90 days | 5.58% | 4.58% | -1.00pp |
| 180 days | 5.84% | 5.17% | -0.67pp |
| 365 days | 6.38% | 6.38% | 0.00pp |

The 14-day staking cohort — primarily liquid staking protocols like BENQI (sAVAX) — stakes at this duration as a structural requirement to service user redemptions, not as a low-commitment choice. These validators run the same infrastructure, maintain 90%+ uptime, and provide the same consensus security as 365-day validators. The short duration exists to serve DeFi liquidity, not to game the reward curve.

At 25M AVAX staked through sAVAX, this proposal costs liquid staking ~320K AVAX/year in lost yield. The cascading effects — reduced competitiveness vs. other chain LSTs, TVL outflows, looping strategies that will break, liquidations on lending markets, weakened DeFi composability — are counterproductive to Avalanche's ecosystem growth.

---

### Proposed Alternative: Piecewise Linear Reward Curve

Rather than lowering a single constant that compresses the entire short-duration range, we propose a piecewise linear interpolation with a breakpoint at 14 days (the current `MinStakeDuration`):

**For d ≥ 14 days:** Existing formula, completely unchanged.
```
rate(d) = MinRate + (MaxRate − MinRate) × d / MintingPeriod
```

**For d < 14 days:** Steep interpolation from a new floor rate to the existing 14-day rate.
```
rate(d) = FloorRate + (rate(14) − FloorRate) × d / 14
```

Where:
- `FloorRate` = 7.5% (new parameter, same value as ACP-285)
- `rate(14)` = `MinRate + (MaxRate − MinRate) × 14 / MintingPeriod` (the existing 14-day rate under the current formula)
- `MinRate` = 10% (unchanged)
- `MaxRate` = 12% (unchanged)
- `MintingPeriod` = 365 days (unchanged)

This produces the following APY schedule:

| Duration | Current APR | ACP-285 APR | Piecewise APR | vs. Current |
|----------|------------|-------------|---------------|-------------|
| 48 hours | - | 4.00% | 4.18% | -1.14pp |
| 7 days | 5.34% | 4.04% | 4.67% | -0.67pp |
| 14 days | 5.36% | 4.08% | **5.36%** | **0.00pp** |
| 30 days | 5.40% | 4.18% | **5.40%** | **0.00pp** |
| 90 days | 5.58% | 4.58% | **5.58%** | **0.00pp** |
| 180 days | 5.84% | 5.17% | **5.84%** | **0.00pp** |
| 365 days | 6.38% | 6.38% | **6.38%** | **0.00pp** |

The 48-hour-to-365-day spread under the piecewise approach is 2.20% — comparable to ACP-285's target of 2.30% — while preserving zero impact on any staker at 14 days or above.

---

### Why 14 Days Is the Natural Breakpoint

14 days is the current `MinStakeDuration` and has been since mainnet launch. It is the boundary between:

1. **Existing proven behavior** — validators have been operating at this cadence since genesis, providing consistent network security.
2. **New behavior ACP-273 enables** — ultra-short 48-hour rolling stakes that have no operational track record and whose security properties are unproven.

The penalty should target the new risk surface ACP-273 opens, not retroactively punish the existing staker base.

---

### Implementation

This proposal is more involved than ACP-285's single-constant change — we want to be upfront about that. The implementation requires:

- A conditional branch in the reward calculator (`vms/platformvm/reward/calculator.go`) to select the appropriate rate formula based on stake duration
- Two new genesis parameters (`FloorConsumptionRate`, `LegacyMinStakeDuration`) threaded through the `Config` struct and calculator constructor
- Updates to three genesis files (mainnet, fuji, local)
- Test coverage for both branches of the conditional

The core logic is:

```
if stakedDuration < LegacyMinStakeDuration:
    legacyRate = MinRate + (MaxRate - MinRate) × LegacyMinStakeDuration / MintingPeriod
    rate = FloorRate + (legacyRate - FloorRate) × stakedDuration / LegacyMinStakeDuration
else:
    rate = MinRate + (MaxRate - MinRate) × stakedDuration / MintingPeriod
```

Realistically this is ~30–50 lines of code versus ACP-285's 3-line diff. No new state transitions, P-Chain transaction types, or consensus logic changes are required. The tradeoff is a modest increase in implementation effort in exchange for completely eliminating impact on every existing staker at 14 days or above.

---

### Addressing ACP-285's Open Questions

**Q1: Is 7.5% the right value?**
7.5% is reasonable as a floor for ultra-short stakes. The piecewise approach makes this question less sensitive — because the penalty is isolated below 14 days, even an aggressive floor (e.g., 5%) would not affect existing stakers.

**Q2: Should MaxConsumptionRate change?**
No. The piecewise approach makes this unnecessary since the existing curve above 14 days is untouched.

**Q3: Should this co-activate with ACP-273?**
Yes. We agree fully. The piecewise curve is specifically designed around the duration range ACP-273 introduces.

---

### Summary

| | ACP-285 | Piecewise Alternative |
|---|---------|----------------------|
| 48hr APR | 4.00% | 4.18% |
| 14-day APR | 4.08% (-24%) | 5.36% (unchanged) |
| 365-day APR | 6.38% | 6.38% |
| 48hr→365d spread | 2.30% | 2.20% |
| Impact on existing stakers | All sub-365d stakers cut | Zero |
| Implementation complexity | Single constant | One conditional |

Both approaches solve the same problem. The piecewise alternative does so without penalizing the infrastructure that is critical to Avalanche's DeFi ecosystem.

ACPs/285-reduce-minimum-consumption-rate/README.md
meaghanfitzgerald: Thanks for the correction!

treehousegovernance:
## Treehouse's Response to ACP-285 and @hinkavax's Piecewise Proposal

**Posted on behalf of the Treehouse team**

[Treehouse](https://app.treehouse.finance/vault/tavax) is a DeFi protocol building LST vaults on Avalanche, currently ranked **No. 19 among all DeFi protocols on Avalanche** by TVL. We would like to formally register our support for the piecewise approach proposed by @hinkavax above, and offer an additional parameter recommendation for consideration.

---

### 1. We Support the Goals of ACP-285

We agree with the core motivation of ACP-285: the combination of ACP-273 (reducing minimum stake duration to 48H) and ACP-236 (auto-renewal) creates a structural incentive misalignment where short-term rolling stakes become nearly economically equivalent to long-term commitments. This weakens the network's validator security posture and should be corrected.

---

### 2. Concerns Around a Simple MinConsumptionRate Reduction

We also echo @hinkavax's concern that a straightforward reduction of `MinConsumptionRate` from 10% to 7.5% — without structural changes to the reward formula — risks causing significant unintended harm to Avalanche's DeFi ecosystem.

The most immediate impact would fall on **Liquid Staking Tokens (LSTs)**, which are a foundational primitive for DeFi on Avalanche. Protocols like BENQI's sAVAX are structurally constrained to stake at or near the 14-day minimum duration in order to facilitate user redemptions. Under the proposed flat reduction, sAVAX's average staking yield would compress to just slightly above 4% — a material deterioration that cannot be offset through protocol design changes alone.

The downstream consequences extend further:

- **Looped LST positions will break.** A significant portion of TVL across Avalanche's lending protocols is built on leveraged LST loops. These strategies are profitable only within a narrow yield band. A ~1.3pp reduction in underlying LST yield would push the majority of these positions into negative PnL territory, triggering forced unwinds.
- **TVL outflows will cascade.** Reduced LST yields → lower LST demand → reduced collateral utilization in lending protocols → reduced DeFi protocol revenues across the board across lending, liquid staking, and vault curation protocols. The unwinding would propagate across the DeFi stack simultaneously, with protocols like Treehouse directly exposed to the second-order impact.

To provide more context on why this matters at an ecosystem level: Treehouse operates as a large-scale but risk-managed looping strategist, performing leveraged LST strategies in size across major Avalanche lending protocols. In doing so, we help tie the spread between staking rates and borrowing rates, contributing to more efficient DeFi capital markets on Avalanche. A severe compression of LST yields directly impairs this function, with unprofitable strategies lowering the lending market utilizations and reducing overall capital efficiency across the ecosystem.

Beyond the mechanical impact, Treehouse also serves as a point of integration for significant institutional and protocol capital on Avalanche. AVAX One has publicly deposited into tAVAX, and we maintain active protocol integrations with Silo and Spectra among others. A material deterioration in tAVAX strategy yield would not only affect our depositors directly — it would create ripple effects across every protocol integrated with tAVAX, potentially triggering coordinated capital outflows that extend well beyond Treehouse itself.

---

### 3. We Support the Piecewise Linear Approach

The piecewise formula proposed by @hinkavax elegantly solves this problem by introducing a breakpoint at 14 days (the current `MinStakeDuration`):

**For `d ≥ 14 days`** — existing formula, completely unchanged:

```
rate(d) = MinRate + (MaxRate - MinRate) × d / MintingPeriod
```

**For `d < 14 days`** — steep interpolation from `FloorRate` to the existing 14-day rate:

```
rate(d) = FloorRate + (rate(14) - FloorRate) × d / 14
```

This approach:
- Leaves the ≥14D reward curve entirely intact, protecting LST yields and the DeFi ecosystem built on top of them
- Creates a steep disincentive specifically for sub-14D staking — the behaviour introduced by ACP-273 and ACP-236 that this ACP seeks to discourage
- Introduces only one new parameter (`FloorRate`) rather than restructuring the entire reward function

---

### 4. Additional Recommendation: Increase MaxRate to 13%

In addition to supporting the piecewise structure, we recommend consideration of increasing `MaxConsumptionRate` from the current **12% to 13%**.

The rationale is twofold:

**Incentive gradient.** With the introduction of 48H staking, the spread between the shortest and longest stake durations should widen to make long-term commitment meaningfully more attractive. Simply penalising the short end is only half the equation — rewarding the long end more explicitly sends a stronger complementary signal.

**Emission sustainability.** With a new class of short-duration validators entering the network post-ACP-273, the network's aggregate stake duration will structurally shorten. Raising `MaxRate` allows the network to sustain a similar level of absolute AVAX issuance to all validators as it does today.

The combined proposal (piecewise formula + `MaxConsumptionRate` raised to 13%) produces the following yield profile versus current parameters:

| Duration | Current APY | Proposed APY | Δ |
|---|---|---|---|
| 48H | 5.32% | 4.18% | -1.14pp |
| 14D | 5.36% | 5.38% | +0.02pp |
| 365D | 6.38% | 6.91% | +0.53pp |

The 14D rate is effectively unchanged (+0.02pp), protecting LST protocols and the DeFi strategies built on top of them. The 48H rate drops by 1.14pp, meaningfully disincentivising short-term rolling. The 365D rate rises to 6.91%, strengthening the long-term staking incentive. Leveraged LST looping strategies will also benefit from the improved long-duration yield, further deepening Avalanche's DeFi ecosystem rather than disrupting it.

---

### Summary of Proposed Changes

| Parameter | Current | Proposed |
|---|---|---|
| `MinConsumptionRate` | 10% | 10% *(unchanged)* |
| `MaxConsumptionRate` | 12% | **13%** |
| `FloorRate` *(new)* | — | **7.5%** |
| Reward formula | Single linear | **Piecewise linear, breakpoint at 14D** |

| Duration | Current APY | Proposed APY | Δ |
|---|---|---|---|
| **48H** | 5.32% | 4.18% | **-1.14pp** |
| **14D** | 5.36% | 5.38% | +0.02pp *(effectively unchanged)* |
| **365D** | 6.38% | 6.91% | **+0.53pp** |

We believe this combination achieves the security objectives of ACP-285 without compromising the integrity of Avalanche's DeFi ecosystem, and we encourage the community to consider it as the basis for the final implementation.

— The Treehouse Team

Nuttymoon:
I agree with @hinkavax and @treehousegovernance comments re: foreseen negative impacts for the DeFi ecosystem.

ACPs/285-reduce-minimum-consumption-rate/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+This is a non-backwards compatible change to P-Chain reward calculation. It requires a network upgrade to activate.
+
+The change affects only new staking periods initiated after activation. Validators and delegators with active staking positions at the time of activation continue to earn rewards under the original parameters for the remainder of their existing stake periods. No migration or re-registration is required.
+
+## Reference Implementation
+
+The complete change is a single floating-point literal update in three genesis files, as specified above. No changes are required to `vms/platformvm/reward/calculator.go`, `vms/platformvm/reward/config.go`, or any other reward logic.
+
+The reward formula itself is not modified. The implementation does not introduce new code paths, state transitions, or P-Chain transaction types.
```
DracoLi: > The implementation does not introduce new code paths

I don't think this is accurate. We can't just change `MinConsumptionRate` in place - doing so would be consensus breaking.

To do this safely we need to keep the existing rate and add the new one, then pick between them based on whether the upgrade is active at the time the reward is calculated.

So the reward formula doesn't change, and there are no new tx types or state transitions - but this isn't a one-line genesis edit.

ACPs/285-reduce-minimum-consumption-rate/README.md
DracoLi: I believe this is correct. The reward is calculated and locked in when the staker is added, not when it's paid out - so anyone with an active position keeps their original rate, and only stakers added after activation get the new one.

ACPs/285-reduce-minimum-consumption-rate/README.md
meaghanfitzgerald: This has been updated in the latest version of the specification

hinkavax:
Posted on behalf of the BENQI Community

Thanks @meaghanfitzgerald, the revision is a clear step forward. The 30-day ramp, grandfathering active stakes, and especially the explicit acknowledgment of risk to LSTs and "relevant strategies" all show the concern is being taken seriously.

We'd build on exactly that LST acknowledgment, with one structural point.

The ramp smooths when the cut lands, not where the 14-day rate ends up. After the ramp, a 14-day stake still settles at ~4.08% vs 5.36% today, a permanent −1.28pp that flows straight into LST yields, lending-market rates, and the spread that keeps looping positions solvent.

More importantly, the flat cut doesn't target the behavior ACP-273 actually enables. The new behavior is sub-14-day rolling stakes — nothing sits below 14 days today, since that's the current MinStakeDuration. Yet under the proposed cut, 48 hours (4.00%) and 14 days (4.08%) are still only ~0.08pp apart, so rolling 48h instead of committing to the structural minimum is barely penalized. The cut mostly lowers the whole floor, and its headline effect is widening the 14d→365d premium from 1.02 to 2.30pp.

A fixed 14-day breakpoint targets the regression directly:

d ≥ 14 days: existing formula, completely unchanged.
d < 14 days: steep interpolation from the 7.5% floor up to the existing 14-day rate.
(Pinned at a fixed 14 days — the pre-273 minimum — deliberately: since ACP-285 co-activates with ACP-273, a breakpoint defined on the live MinStakeDuration would collapse to 48h at activation and the steep segment would cover nothing.)

This produces a real ~1.2pp penalty between 48h (4.18%) and 14d (5.36%), a far sharper disincentive against sub-14-day stakes than the flat cut's ~0.08pp, while preserving the 14-day rate entirely. It needs no transition ramp of its own, because it disturbs no existing stake: everything staked today is at or above 14 days and stays untouched.

It's also worth being concrete about who would actually occupy the sub-14-day range once ACP-273 opens it. The natural occupants are custodial and ETF stakers, Grayscale's GAVA and Bitwise's BAVA already hold and stake AVAX, whose binding constraint is redemption liquidity, not yield: staked AVAX they can't unwind quickly impairs their ability to meet redemptions. That cohort will gravitate to the shortest available duration and is largely indifferent to the rate, accepting a lower 48h yield as the price of near-daily liquidity. The steep sub-14-day segment therefore does double duty, a genuine disincentive for discretionary stakers who could commit longer, and an appropriate liquidity discount for the ETF cohort that they'll pay rather than be deterred by. The flat cut, by contrast, lowers the 14-day rate paid to yield-sensitive LSTs that feed DeFi: it lands on the cohort that minds and spares the one that doesn't.

On the duration premium: the breakpoint leaves 14d→365d at 1.02pp rather than widening it, and we'd argue that's correct. Widening that premium is the speculative lever, the proposal's own Goal 3 projects only a ~2-month shift in stake-weighted duration and frames it as an expected secondary effect, not a guaranteed one, with mixed evidence on whether large validators move at all. Sacrificing the structural 14-day rate (at the cost of real DeFi losses) to widen a premium the model says may not change behavior is a poor trade.

On inflation, the breakpoint still trims issuance to the sub-14-day cohort; it forgoes only the broad level reduction across the structural 14–365 day range. Weighing that honestly: the reduction the breakpoint gives up is diffuse (spread across all holders as marginally-lower dilution), modest, and reached only gradually, whereas the harm the flat cut imposes is concentrated on the full-security DeFi cohort, effectively certain, and amplified by second-order effects in looping and lending. We'd rather not fund a diffuse, gradual emissions reduction by compressing the rate paid to the validators who provide both security and DeFi liquidity, particularly when the breakpoint preserves most of the structural benefit and lands the cut on the cohort that accepts it.

Ask: adopt a fixed 14-day breakpoint, concentrating the gradient where ACP-273 creates the regression. If a curve change is out of scope for the Helicon timeline, extend the ramp from 30 to 60–90 days as an interim cushion.

avalanche-foundation-admin (approved):

coad1024-cmd:
Hi all,

I am Hash, Research Engineer at the [Bonding Curve Research Group (BCRG)](https://x.com/Bonding_Curves). BCRG specializes in the design, analysis, and operationalization of socio-technical systems. Our core focus is equipping protocols with the data-driven insights needed to validate tokenomics and calibrate policy controls amidst market uncertainty.

Currently, we are advancing our collaboration with the Avalanche Foundation. Having established the foundational taxonomies, systems analysis, and differential specification in our [Phase I research](https://bonding-curves.github.io/avalanche-research/), we are now initiating Phase II. This next stage centers on a rigorous investigation of the Avalanche staking subsystem using the Parameter Selection Under Uncertainty (PSUU) methodology. To ensure these models align with real-world dynamics, we will be conducting targeted stakeholder interviews to directly inform and drive upcoming policy decisions on the network.

---

In the meantime, I wanted to share some initial simulation results from my own quantitative modeling of P-chain staking dynamics to help inform the immediate discussion around ACP-285.

I want to offer a first-principles alternative to both the flat floor cut (proposing a 7.5% linear floor) and the piecewise linear 'kink' at 14 days. 

While the piecewise linear curve is a strong candidate for protecting LST yields, hardcoding "14 days" into consensus-critical code introduces structural debt. More importantly, **my quantitative modeling shows that a piecewise curve fails to solve the inflation reduction goal.**

Below, I compare the **Piecewise Linear Curve** against a continuous, **Asymmetric Polynomial Bump Curve** using simulations from a cadCAD model calibrated on historical P-chain staking data.

---

### The Mathematical Conflict of a Piecewise Curve

Under the proposed piecewise curve:
* For $d \ge 14$ days: The rate remains identical to the original status quo linear rate (starting at 10% and scaling to 12%).
* For $d < 14$ days: The rate scales linearly from a 7.5% floor to the 14-day rate.

While ACP-273 proposes shortening the minimum staking duration to 2 days (48 hours), the vast majority of network-wide staking supply (institutional validators, long-term holders, and LST protocols like Benqi) is expected to continue locking capital for 14 days or longer. This is because short 2-day locks incur substantial transaction and node renewal overhead while paying a lower rate.

Because the piecewise curve immediately reverts to the high status quo rates for all durations $\ge 14$ days, the bulk of the staking supply receives no rate cut. As a result, **the network-level average inflation remains completely unchanged at 2.19%**, failing the primary inflation reduction objective of ACP-285.

### The Solution: An Asymmetric Polynomial Bump Curve

Instead of a piecewise kink, I propose starting with the 7.5% linear floor and adding a continuous, asymmetric "bump" that peaks at 14 days and decays rapidly to zero for longer durations:

$$C(d) = 7.5\% + 4.5\% \cdot \left(\frac{d}{365}\right) + \lambda \cdot \left(\frac{d}{365}\right) \cdot \left(1 - \frac{d}{365}\right)^{25}$$

Where:
* $d$ is the staking duration in days.
* $\lambda = 127.4$ is a scaling factor that calibrates the height of the bump.
* The exponent $25$ shifts the peak of the bump to exactly $\frac{1}{26}$ of the year ($\approx 14$ days).

---

### Expected Yields (Consumption Rates & Realized APYs)

At a network current supply of 472.3M AVAX, the scarcity factor is $\approx 0.524$. The consumption rates and final realized APYs compare as follows:

| Duration | Current (10% - 12%) | ACP-285 (7.5% - 12%) | Piecewise (7.5% - 12%) | **Asymmetric Bump (7.5% - 12%)** |
| :--- | :---: | :---: | :---: | :---: |
| **48 Hours (2d)** | 10.01% *(5.25% APY)* | 7.52% *(3.94% APY)* | 7.87% *(4.12% APY)* | **8.13% *(4.26% APY)* (Short lock penalty)** |
| **14 Days** | 10.08% *(5.28% APY)* | 7.67% *(4.02% APY)* | 10.08% *(5.28% APY)* | **9.50% *(4.98% APY)* (DeFi Protected)** |
| **30 Days** | 10.16% *(5.32% APY)* | 7.87% *(4.12% APY)* | 10.16% *(5.32% APY)* | **9.10% *(4.77% APY)* (Decaying)** |
| **90 Days** | 10.50% *(5.50% APY)* | 8.61% *(4.51% APY)* | 10.50% *(5.50% APY)* | **8.63% *(4.52% APY)* (Fully decayed)** |
| **365 Days** | 12.00% *(6.29% APY)* | 12.00% *(6.29% APY)* | 12.00% *(6.29% APY)* | **12.00% *(6.29% APY)* (Ceiling intact)** |

---

### cadCAD Simulation Results (1,600 Config Sweep)

To evaluate the policy robustness, I leveraged the **Parameter Selection under Uncertainty (PSUU)** methodology. By sweeping 1,600 configurations across various governance parameters (curves, caps, opportunity costs) and environmental assumptions, we can evaluate curve behaviors under stress.

*(Note: The full cadCAD simulation codebase will be open-sourced in the coming months.)*

| Policy Arm | Curve Shape | Mean Staking Ratio ($\theta$) | Mean Inflation Rate |
| :--- | :--- | :---: | :---: |
| **`status_quo`** | Linear (10% - 12%) | **45.77%** | **2.19%** |
| **`acp285_filed`** | Linear (7.5% - 12%) | **44.48%** | **1.81%** |
| **`piecewise_compromise`** | Piecewise Linear | **45.77%** | **2.19%** *(No inflation reduction)* |
| **`asymmetric_bump`** | Asymmetric Bump (7.5% - 12%) | **44.74%** | **1.89%** *(~75% of ACP-285 savings)* |

---

### Staker Preference Modeling (Discrete-Choice Logit under Eric Lu's Betas)

A common critique of curve reshaping is that compressing short-term yields might cause stakers to lock up tokens for shorter periods. To evaluate this behavioral margin, I implemented a **discrete-choice logit model** of commitment duration.

Staker utility $U(d)$ is defined as:
$$U(d) = \text{ASC}_d + \beta_{\text{apy}} \cdot \ln(\text{APY}(d)) + \beta_{\text{dur}} \cdot \ln(\text{weeks}(d))$$

I calibrated the model using the **delegator coefficients estimated by Eric Lu** in his published staking research (Appendix A: $\beta_{\text{apy}} = +0.444$, $\beta_{\text{dur}} = -0.643$) and matched it against the P-chain’s short-skewed baseline duration distribution (where over 85% of stake is locked for $\le 30$ days).

Swapping in the alternative curve parameters yielded the following choice shifts:

| Policy Arm | Mean Staking Lock | Staking Lock $\ge 90$ Days | Change vs. Status Quo (Days) |
| :--- | :---: | :---: | :---: |
| **`status_quo`** | 39.50 days | 14.00% | *Baseline* |
| **`acp285_filed`** | 40.82 days | 14.65% | **+1.32 days** (Lengthened lockup) |
| **`piecewise_compromise`** | 39.50 days | 14.00% | **+0.00 days** (No behavioral shift) |
| **`asymmetric_bump`** | 39.35 days | 13.65% | **-0.15 days** (Inert / Neutral) |

#### Takeaways from Preference Modeling:
1. **The Piecewise Curve is behavioral noise:** Because the rates stakers face for all valid durations $\ge 14$ days remain identical to the status quo, staker utility is unchanged. Stakers make the exact same choice distribution as the status quo ($+0.00$ days shift).
2. **The Asymmetric Bump Curve is inert on duration choice:** It shifts mean commitment duration by a negligible **-0.15 days** (less than 4 hours). This demonstrates that smoothing the curve to reduce inflation does not trigger duration degradation or validator set churn.
3. **Active controls are level and cap, not shape:** These results align with the finding that curve *shape* within today's narrow band is inert on chosen duration. The only active lever that successfully drives stakers to commit longer is lowering the reward floor (as ACP-285 does, shifting lockups by $+1.32$ days).

---

### Key Policy Takeaways:

1. **The Piecewise Curve is a No-Op on Inflation:** Because all valid stakers must lock for $\ge 14$ days, they sit entirely on the un-modified status quo portion of the piecewise curve. Thus, the piecewise curve results in a mean inflation rate of **2.19%** (exactly identical to the status quo). It fails the primary objective of ACP-285.
2. **The Asymmetric Polynomial Bump Curve succeeds on both fronts:** It decays slowly and smoothly over the 14-to-90 day range. This allows it to protect the 14-day LST rate at **9.50%** (yielding ~4.98% realized APY to keep staking participation stable at **44.74%**), while still structurally reducing inflation for longer durations to **1.89%** (capturing ~75% of the total desired inflation savings).
3. **Consensus Safety:** Bending curves on blockchains is risky because floating-point calculations are non-deterministic. Because the asymmetric bump curve uses a fixed integer exponent ($25$), it can be computed using simple, loop-based integer multiplication. It is computationally trivial, easy to audit, and carries zero consensus fork risk.

### Visualizations

#### 1. cadCAD Staking Subsystem Simulation (1,600 Config Sweep)
The four-panel plot below demonstrates that the `piecewise_compromise` curve (yellow) overlaps perfectly with the `status_quo` (blue) across all distributions, while `asymmetric_bump` (red) successfully compresses inflation while maintaining participation.
oq2_four_panel


#### 2. Staker Duration Choice Model (Discrete Logit under Published BLP Betas)
The plot below illustrates the counterfactual choice shares and changes in mean commitment duration under each curve design.
duration_logit

---

hinkavax:
Thanks @coad1024-cmd. Your proposal is excellent, and have our full support.

treehousegovernance:
Excellent proposal @coad1024-cmd. The Treehouse team supports this implementation and hopes the community takes it into consideration for ACP-285.

ericlu-avax:
@coad1024-cmd Thank you for the thorough analyses. The results look very interesting.
Could you provide more clarity on the following?

1. The exact math formula for the staking reward rate curve, i.e., annual reward rate as a function of staking duration.
2. The asymmetric bump curve appears to lead to (weakly) higher ARR for validators at all staking durations compared to the the straight line curve in the current protocol. Yet, we get to see 75% of the inflation rate drop. Could you help me understand where the drop in inflation rate comes from while the validators seem to have higher ARR across the board?
3. Could you also look into the uncertainty on the effects on inflation rate? i.e., what is the distribution of inflation rate with the asymmetric bump curve vs the straight line?

PR #287, opened 2026-05-28
Formalize ACP-283

StephenButtolph:
This doesn't change any of the ACP motivations, but adds significant details required to correctly implement the ACP.

Some key changes that are specified here:
1. The ACP no long includes a hard-cap on the minimum gas fee. This was only briefly mentioned in the `Security Considerations` section. But is now removed.
2. The ACP now specifies that $P$ is expected to be continuous when changing $M$ and $M' \leq P$.
3. The ACP includes details required to integrate into SAE (which will eventually be used with subnet-evm, so the construction is made to operate easily with a dynamic minimum price that isn't exponential and voted on by validators)
4. The value for $D$ is formulated to allow the highest level of specificity without incurring any performance overhead
5. The value for $Q$ is defined based on the human readable number of blocks to double.
6. (May be controversial) the number of blocks to double is set to 3,600. I still need to inspect how low this number can go and have it still be "safe"... But I think this is a reasonable placeholder (currently it isn't specified). We can update $Q$ in a followup if needed.


powerslider:
Based on some initial remarks from @StephenButtolph I've modified the implementation here: https://github.com/ava-labs/avalanchego/pull/5432. Just citing this here for cross reference checking to be sure that these spec modifications comply with the implementation part.

martineckardt (approved):

avalanche-foundation-admin (approved):

PR #289, opened 2026-06-17
[ACP-285] Increase 30 day reduction period to 90 day reduction period

meaghanfitzgerald:
Extends the MinConsumptionRate reduction ramp from 30 to 90 days. This lengthens the transition window, smoothing the per-day rate change during the ramp (~0.083pp → ~0.028pp) and giving LSTs and DeFi integrators more time to reprice. The 7.5% floor, the ceiling, and all final ARR values remain unchanged. 
 
Made in response to BENQI Community feedback, which proposed extending the ramp to 60–90 days as an interim cushion: https://github.com/avalanche-foundation/ACPs/pull/285#issuecomment-4670008219

### DIFF
--- a/ACPs/285-reduce-minimum-consumption-rate/README.md
+++ b/ACPs/285-reduce-minimum-consumption-rate/README.md
@@ -114,9 +114,9 @@ $$r(t) = r_0 - \Delta \cdot \frac{\min(t - t_0,\ P)}{P}$$
 
 This decreases linearly from $r_0$ at $t_0$ to $r_0 - \Delta$ at $t_0 + P$, and holds at $r_0 - \Delta$ thereafter.
 
-$r(t)$ is evaluated once per stake, using the stake's start time, and is fixed for that stake's full duration. A stake beginning during the ramp locks in the rate at its start, so a stake that starts on day 15 uses $8.75\%$ for its entire term. Stakes already active at $t_0$ are unaffected, since their reward is determined at their own start time.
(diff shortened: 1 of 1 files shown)


avalanche-foundation-admin (approved):

hinkavax:
Posting on behalf of the BENQI community:

Thank you @meaghanfitzgerald for extending the reduction period to 90 days.

We want to be clear, though, that it doesn't resolve the issue, because the harm was never primarily about the speed of the transition. It's about where the 14-day rate lands. A 90-day glide and a 30-day glide arrive at the same destination, a permanent 14-day rate of ~4.08% versus 5.36% today, a ~1.28pp (~24%) reduction that holds indefinitely. That terminal level is the specific problem for leveraged strategies. A permanent 1.28pp cut to the base yield could lead to a material share of looped positions to unwind.

Which brings us to the question we'd most like an answer to: what justifies the full 2.5pp reduction to 7.5%, specifically, against the proposal's own behavioral projection? Goal 3 estimates the change shifts stake-weighted duration by only ~2 months, frames it as an expected secondary effect rather than a guaranteed one, and acknowledges mixed evidence on whether large validators move at all. If the purpose of widening the duration premium is to lengthen staking, and the modeled gain is ~2 months with low confidence, it is hard to see why a reduction of the full magnitude, with its certain, permanent cost to LSTs and the DeFi built on them, is the right size.

We still regard the 14-day breakpoint as the cleaner resolution.

